Issue #2, opened 2020-06-30
Figma source access request


jimador-sf:
Requesting access permission to the Figma source link in the test.

marcin-piela-sf:
@jimador-sf you should have access to it

https://www.figma.com/file/hd7EgdTxJs2fpTzzSKlNxo/Superformula-full-stack-test

Just create an account and you will be able to read/check css styles. You don't need write access to that file.

jimador-sf:
Thank you!

Issue #1, opened 2020-06-30
GQL Mutation requirements


jimador-sf:
From the requirements, and Figma mockups, it appears that the only mutations required are `createUser` and `updateUser`. Is the delete functionality required?

afaq-sf:
Great question.

Only behavior depicted in Figma or mentioned in the Readme is required. 

If you notice additional system-level additions that add value to the product, we would welcome them as additional bonus contributions, whether as code or just a description of how you'd go about it. We would then use it as a launching off point for discussion during the PR review.

Please let us know if that provides enough clarity.

jimador-sf:
@afaq-sf Sounds great. I added a **_Gold-plating_** section to the kanban board I linked in the updated README

PR #7, opened 2020-07-03
Karlgolka

charlieargue:
Hi All!

Please see relevant readme files under the `fe` and `be` front-end and back-end sub-folders!

You'll need .env files for each, I will email you my two API keys/URI for now so you can easily spin all this up locally if-you-so-choose, you can use those temporarily!

Thanks, Happy 4th of July!


be/graphql/resolvers.graphql.js
```diff
@@ -0,0 +1,100 @@
...
+const User = require('../models/user')
+
+const resolvers = {
+    Query: {
+        allUsers: () => {
+            return User.find({})
+        },
+        filteredUsers: async (root, args) => {
+            try {
+                const perPage = 6
```
afaq-sf, line 11: In a real-world implementation, how would you recommend making this pagination property configurable by the API team?

be/graphql/resolvers.graphql.js
```diff
@@ -0,0 +1,100 @@
+const { UserInputError } = require('apollo-server')
+const User = require('../models/user')
+
+const resolvers = {
+    Query: {
+        allUsers: () => {
+            return User.find({})
+        },
+        filteredUsers: async (root, args) => {
```
afaq-sf, line 9: What are your thoughts on making `args` a destructured object?

For example: `async (root, {page = 0, search = false,...})`

be/graphql/resolvers.graphql.js
```diff
@@ -0,0 +1,100 @@
...
+            }
+        },
+        editUser: async (root, args) => {
+            const user = await User.findOne({ _id: args.id })
+            user.name = args.name
+            user.address = args.address
+            user.description = args.description
+
+            try {
+                return await user.save()
```
afaq-sf, line 80: Would you consider using `findByIdAndUpdate()` mongoose method here instead?

be/graphql/resolvers.graphql.js
```diff
@@ -0,0 +1,100 @@
...
+            } catch (error) {
+                throw new UserInputError(error.message, {
+                    invalidArgs: args,
+                })
+            }
+        },
+        deleteUser: async (root, args) => {
+            const user = await User.findOne({ _id: args.id })
+            try {
+                return await user.remove()
```
afaq-sf, line 90: In a real-world app, how would you design for archiving user records rather than truly deleting them?

be/graphql/resolvers.graphql.js
```diff
@@ -0,0 +1,100 @@
+const { UserInputError } = require('apollo-server')
+const User = require('../models/user')
+
+const resolvers = {
+    Query: {
+        allUsers: () => {
```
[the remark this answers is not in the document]
charlieargue: Yes, of course (and thanks!).

This is “legacy code“ in a way – I first started with fulfilling the basic CRUD app requirements, and wanted to provide easy graphQL methods for each C.R.U.D. operation (a simple read in this case), like I did at the end of the be/README.md (since careful and clear documentation was also a requirement).. 

Then I later added search and paging functionality, replacing the allUsers query entirely, but I kept it around as a simple example of a read operation, for documentation purposes (to quickly throw in the playground, etc). 

In production, you’re absolutely right – this method could be removed and it’s functionality can easily be retained by making the filteredUsers params optional and have defaults.

P.S. - If I may also add, if I were to work on this another day, the first upgrades I would add would be upgrading to redux and switching to Dynamo and AppSync (now that AWS finally activated my account... details are in my email to Will... so apologies for not using AWS!).

be/graphql/resolvers.graphql.js
charlieargue: I typically have ended up putting these in **.env files** as a KEY, but you could just load it from a **global constants file**. 

For a **more robust option**, you can have a `system_setting` collection or table and store it in there (thereby making it easier to, for example, save individual user-specific settings, i.e. I have a large monitor and want to load 60 users, etc.) – or both, a .env easily configurable  default (or stored in system_setting), and then perhaps you also allow users to customize their default (so not machine specific, but whenever the user logs in they would get their preferred paging amount). By just including a userId column that could be null for a system-wide default.

be/graphql/resolvers.graphql.js
charlieargue: I think that’s a great idea, thanks – clearer for sure!

In my **state variables**, I initialized search to null, if I recall.   
```
const [page, setPage] = useState(0)  
const [search, setSearch] = useState(null)      
```

So now these initialization values need to remain synced to keep things clear.

Aside: in my node/express API at LJI, I would use Joi validation files for each endpoint -- and this was one place where defaults could be set.

be/graphql/resolvers.graphql.js
charlieargue: Absolutely! Forgot about that one. It’s clearer and can be done in one command, easily passing an entire changeset.    

However, I would want to **set the option** `new: true`, so that it returns the newly updated object (by default, apparently, it returns the old object).

be/graphql/resolvers.graphql.js
charlieargue: Yes, certainly – at LJI I would **soft-delete** all the primary records and their children instead of hard-deleting them (carts, orders, requisitions, etc.) – a feature the ORM I used provided (Sequelize, as most ORMs do). 

(Then by default, all ORM queries were against non-soft-deleted records,and if you wanted to query the soft-deleted records, you just always added `where: deletedAt` is not null... to your queries). 

And of course, you can actually hard-delete records as well at any time.

charlieargue: Additionally, some records were never soft-deleted, such as user activity records, etc.

fe/src/components/UserForm.js
```diff
@@ -0,0 +1,191 @@
...
+            setDescription('')
+            // reset showMap
+            setShowMap(false)
+
+            // and close it
+            onCancel()
+        }
+    })
+    // --------------
+    const submit = async (event) => {
```
marcin-piela-sf, line 62: Have you considered using form library like formik/react-hook-form? Why or why not?

fe/src/components/UserForm.js
```diff
@@ -0,0 +1,191 @@
...
+                        <Form noValidate validated={validated} onSubmit={submit}>
+                            <Form.Group controlId="formName">
+                                <Form.Label>Name</Form.Label>
+                                <Form.Control value={name || ''}
+                                    onChange={({ target }) => setName(target.value)}
+                                    type="text"
+                                    required
+                                    placeholder="Enter name" />
+                                <Form.Control.Feedback type="invalid">
+                                    Please enter a name
```
marcin-piela-sf, line 147: In a real-world implementation, how would you deal with texts in application?

fe/src/components/UserList.js
```diff
@@ -0,0 +1,74 @@
+import React, { useState, useEffect } from 'react'
+import { useQuery } from '@apollo/client'
+import { FILTERED_USERS } from '../queries/queries.graphql'
```
marcin-piela-sf, line 3: Have you considered using global imports? Why or why not?

fe/src/components/UserList.js
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [result, getSearchPhrase, setMessage, setShowNotification])
+
+    // --------------
+    if (result.loading) {
+        // TODO: style me!
+        return <div>loading...</div>
```
marcin-piela-sf, line 53: Would you consider using `Spinner` component from `react-bootstrap` here?

fe/src/fonts/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2010, 2012, 2014 Adobe Systems Incorporated (http://www.adobe.com/), with Reserved Font Name ‘Source’.
```
marcin-piela-sf, line 1: Would you consider using https://fonts.google.com/ to serve fonts in real world app instead of adding them to project? Why or why not?

fe/src/globals.js
```diff
@@ -0,0 +1,6 @@
+const GLOBALS = {
+    API: {
+        BASE_URL: 'http://localhost:4000'
```
marcin-piela-sf, line 3: What are your thoughts on moving it to .env?

fe/src/queries/queries.graphql.js
```diff
@@ -0,0 +1,51 @@
+import { gql } from '@apollo/client';
+
+// GRAPHQL queries
+export const ALL_USERS = gql`
+query {
+    allUsers {
+        id
+        name,
+        address,
+      	description,
```
marcin-piela-sf, line 10: Would you consider using eslint/prettier to fix and unify spacing etc across all files?

fe/src/queries/queries.graphql.js
```diff
@@ -0,0 +1,51 @@
...
+`
+
+  
+
+export const EDIT_USER = gql`
+mutation editUser(
+    $id: ID!,
+    $name: String!,
+    $address: String!,
+    $description: String!
```
marcin-piela-sf, line 35: What will you do when in the future that mutation will require at least 10 params?

fe/src/ui/styles.css
```diff
@@ -0,0 +1,160 @@
...
+  font-weight: 300 !important;
+  font-size: 48px !important;
+  line-height: 60px !important;
+}
+
+.row.row-centered {
+  justify-content: center;
+}
+
+.card {
```
marcin-piela-sf, line 48: In a real-world app, where would you put components styles?

fe/src/components/CenteredModal.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Modal } from "react-bootstrap";
+
+// --------------
+// --------------
+// --------------
+const CenteredModal = (props) => {
```
marcin-piela-sf, line 7: Have you checked the margins / spacing of this component? Is this component the same as on designs?

fe/README.md
```diff
@@ -0,0 +1,68 @@
...
+# change directory into the front-end subdirectory
+cd fe
+
+# install all libraries
+npm i
+```
+
+# Quick Start
+
+You will need to grab a `.env` environment variable file:
```
marcin-piela-sf, line 24: Have you considered adding .env.dist file with example configuration?

fe/src/components/Dashboard.js
```diff
@@ -0,0 +1,93 @@
...
+import { debounce } from 'lodash'
+import { scrollToBottom } from '../services/utilities.service'
+import { useHistory } from "react-router-dom";
+
+// --------------
+// --------------
+// --------------
+const Dashboard = ({ setShowNotification, setMessage }) => {
+    const history = useHistory()
+    const [modalShow, setModalShow] = React.useState(false);
```
marcin-piela-sf, line 17: Why in one place you use "React.useState" and in another "useState"? Is it on purpose?

fe/src/components/layout/ApplicationLayout.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Container, Row, Col } from "react-bootstrap";
+import { withRouter } from "react-router";
```
marcin-piela-sf, line 3: What are your thoughts on importing `Router` component instead of `withRouter` HOC?

fe/src/components/UserForm.js
```diff
@@ -0,0 +1,191 @@
...
+                                    required
+                                    placeholder="Enter description" />
+                                <Form.Control.Feedback type="invalid">
+                                    Please enter a description
+                                </Form.Control.Feedback>
+                            </Form.Group>
+                            <Container fluid className="pl-0 pt-4">
+                                <Row>
+                                    <Col>
+                                        <BigButton type="submit" text="Save" autoFocus id="btnSaveUserForm" />
```
marcin-piela-sf, line 175: How would you block user from submitting same data multiple times?

fe/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "author": "Karl Golka",
+  "license": "MIT",
+  "dependencies": {
+    "@apollo/client": "3.0.0-rc.10",
+    "bootstrap": "4.5.0",
+    "google-maps-react": "2.0.6",
+    "graphql": "15.2.0",
+    "lodash": "4.17.15",
```
mat-sf, line 12: As lodash library is quite big - do you have any ideas on how to optimize it (let's assume that we have our own webpack config, not CRA)?

fe/src/queries/queries.graphql.js
```diff
@@ -0,0 +1,51 @@
...
+      	description,
+        createdAt
+    }
+}
+`
+
+export const FILTERED_USERS = gql`
+query FilteredUsers($search: String, $page: Int) {
+    filteredUsers(search: $search, page: $page) {
+      id
```
mat-sf, line 19: We query the same fields (id, name, description, address, createdAt) in `allUsers` and `filteredUsers` queries - do you have any idea how to unify it and not repeat it in every users queries?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
charlieargue: No, apologies – that was accidental – I have git-removed it and pushed up the change.

fe/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "fe",
+  "version": "0.1.0",
+  "private": true,
+  "author": "Karl Golka",
+  "license": "MIT",
+  "dependencies": {
+    "@apollo/client": "3.0.0-rc.10",
+    "bootstrap": "4.5.0",
```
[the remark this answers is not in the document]
charlieargue: I personally usually prefer Google’s Material design, but given the target UI for this project -- my initial feeling was that the default bootstrap UI lined up closer and would require less customization. Also, I had just finished a material-ui react project, and wanted to try react-bootstrap!

But I really used react-bootstrap (not the old bootstrap) -  https://react-bootstrap.github.io/getting-started/introduction (which requires only the CSS portion of bootstrap, not the jquery or javascript) – react-bootstrap, like ng-bootstrap, is not jquery dependent, and replaces the old bootstrap JavaScript with native components.

I would support using react-bootstrap on production apps, and I think it was a good choice. You can always go for more minimalist design frameworks, tailwind.css, ant, there’s sooooo many good systems out there now, for example many of these:
-	https://github.com/bradtraversy/design-resources-for-developers#react-ui

This makes me want to discuss mobile-first design, lighthouse scores, and responsive/adaptive design, pre-loading images, using skeletons, and lots of other front-end performance choices (that I skipped for this project, needless to say) -- great question!

fe/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "google-maps-react": "2.0.6",
+    "graphql": "15.2.0",
+    "lodash": "4.17.15",
+    "moment": "2.27.0",
```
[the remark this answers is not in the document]
charlieargue: Good question! So my last big project, in angular, there was a way to import a slimmer version of moment (without the localization parts, iirc) – I would look for a way to do something similar here. It is true, that moment is a larger dependency. 

I used https://www.npmjs.com/package/webpack-bundle-analyzer to quickly trouble-shoot dependency bloat in the past, and I remember momentJS being one of the larges bundles in angular!

But you’re absolutely right, that for performant production apps, you don’t need moment! https://github.com/you-dont-need/You-Dont-Need-Momentjs

Something like Luxon, or react-dayjs might be a better way to go, or for just one function (unix timestamp -> user friendly) – just write it oneself.

fe/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+
+
+// prep the preloading promises (kicks off immediately when the current file is imported)
+const promiseDashboard = import("./components/Dashboard.js")
+
+// and preload! (by the time this gets rendered, your component is probably already loaded)
+const Dashboard = lazy(() => promiseDashboard)
+
+// --------------
+// --------------
```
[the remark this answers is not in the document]
charlieargue: Just a visual tick-mark to let me scan quickly (bad habit I picked up as I was learning react) – I removed them, should be reflected in the PR already!

fe/src/App.js
```diff
@@ -0,0 +1,49 @@
...
+// --------------
+const App = (props) => {
+    const [showNotification, setShowNotification] = useState(false);
+    const [message, setMessage] = useState('')
+
+    // -------------- (state variables getters to pass to children and other components)
+    const getShowNotification = () => showNotification
+    const getMessage = () => message
```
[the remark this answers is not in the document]
charlieargue: I’ve removed the unnecessary getters (getShowNotification and getMessage) and am now passing the string state variable `message` and boolean `showNotification` directly – pushed up another commit, my mistake.

fe/src/components/BigButton.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Button } from "react-bootstrap";
+
+// --------------
+// --------------
+// --------------
+const BigButton = (props) => {
+    const styles = {
+        background: '#FFFFFF',
```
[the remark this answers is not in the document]
charlieargue: Yes, absolutely! A more robust approach than what I did here would be to reference the un-compiled bootrap .scss SASS file, and use a full templating and over-riding approach  – using mixins and creating variables for your theme, and being able to over-ride bootstrap styles in a central place (bootstrap 5 will be pretty cool with the new features like $utilities: merge-map()!).

On another react project using material-ui, I had a custom hook that served as a central place for .css styles that integrated with the material-ui theme infrastructure and syntax. 

So I ended up having styles in three places (react material-ui project):
-	1) SASS file _mixins that over-ride the core library styles and reference it’s variables, etc… (a re-usable background-radius class, or over-rides, for example, would live here)
-	2) Styles that exist only specific to each component
-	3) A global styles hook that had app-wide re-usable css styles that had to live AFTER the sass was loaded since they integrated with the main theme object for responsive styling and access to theme spacing, palette colors, etc. (things like application layout responsive styles live here, etc.)

See SASS answer below for more details.

fe/src/components/BigButton.js
```diff
@@ -0,0 +1,23 @@
...
+        border: '4px solid rgba(0, 0, 0, 0.1)',
+        boxSizing: 'border-box',
+        borderRadius: '8px',
+    }
+    // --------------
+    return (
+        <Button {...props}
+            className="px-5 py-3 big-button"
+            style={styles}
+            size="lg">{props.text || "Load More"}</Button>
```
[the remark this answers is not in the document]
charlieargue: > Would you consider allowing to set size in BigButton props?

Absolutely! I think I was going for speed of development, so used the {…} props generic approach, since I wasn’t sure what I would need to expose, it was just a rough sketch, so-to-speak. So yes, exposing the size property could work for a more responsive/adaptive project 

Aside: I wrote a custom react hook on my last project that gave me isPhone, isTablet, isDesktop, isExtraLarge easy-to-use device-sniffer properties, that could be used, just as an example, to determine the size being passed in, something like:
```
<BigButton size={isPhone ? “sm” : isTablet ? “lg” : “xl” } … 
```

> Do you think that Load More is a good default value here?

No! I think I knew I would only use this button three times exactly, and so kept it rough, apologies again! A better default would be something like “Save” or even an empty string, and you would want to pass in <BigButton text=“Load More”… of course.

fe/src/components/UserForm.js
```diff
@@ -0,0 +1,191 @@
...
+
+    // --------------
+    useEffect(() => {
+        debouncedGeocoding(currentUser.address)
+    }, [currentUser.address, debouncedGeocoding])
+
+    // --------------
+    return (
+        <>
+            <Container fluid className="pb-2">
```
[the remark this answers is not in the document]
charlieargue: By using device sniffing (something like in my previous post) to be able to make some granular modifications based on device. 

But mostly be using layout properties like `<Col xs={12} md={6}>` which would make that column full-width (12) for xs devices (aka phones), and medium devices and up would be half-width (6, so two-columns), etc. That, and perhaps also actual CSS `@media queries`, if needed.

fe/src/components/CenteredModal.js
```diff
@@ -0,0 +1,26 @@
...
+    // --------------
+    return (
+        <Modal
+            {...props}
+            size="xl"
+            aria-labelledby="contained-modal-title-vcenter"
```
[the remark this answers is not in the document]
charlieargue: Just copy-paste from the react-bootstrap Modal example I used to start: - https://react-bootstrap.github.io/components/modal/#modal-vertically-centered

Was not doing responsive/accessibility, but kept it in just as an example.

fe/src/components/UserCard.js
```diff
@@ -0,0 +1,50 @@
...
+        <Card className="p-3 py-5 mb-5" onClick={() => openModal(true)}>
+            <Image
+                className="edit-icon"
+                src={EditIcon}
+                style={{ display: "none" }}
+            />
+            <div className="align-self-center mb-2">
+                <Image
+                    style={{
+                        background: `transparent url('https://source.unsplash.com/collection/3053437/${user.id}') no-repeat 50% 50%`,
```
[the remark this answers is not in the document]
charlieargue: > Have you considered moving inline styles to CSS?

 Yes, good eye. I have moved all-but-one of those styles to the .css file:
-	`background: transparent url('https://source.unsplash.com/collection/3053437/${user.id}') no-repeat 50% 50%`,
Kept that one because it has the dynamic `user.id` component.

> Do you think is it a good idea to make Avatar as a separate component?

Yes, absolutely! Then you can **better re-use it through-out the site** – like on a user-profile page, on a list of users elsewhere, etc…

fe/src/components/UserCard.js
```diff
@@ -0,0 +1,50 @@
...
+                </Card.Title>
+                <Card.Text className="text-ellipsis">
+                    {user.description || "Lorem ipsum dolor sit amet, consectetur…"}
+                </Card.Text>
+            </Card.Body>
+        </Card>
+    )
+}
+
+export default UserCard
```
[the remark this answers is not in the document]
charlieargue: Yes, if you want to combine multiple components or utilities into one file. 

Like when we do: `import React, { useState } from 'react'`

`React` is the default export and everything in the `{ brackets }` are the named exports. 

This helps when you want to be able to only bring in what you need and not bloat your dependencies unnecessarily, and still be able to make it easy to grab the default export all from one file. 

For example, we can put all the User Card, Form, and List components together into one export file (“barreling”) – so if it is being used in another project, it is easy to bring in the entire set or as-needed using only one import statement.

I used named exports in my [utilities service](https://github.com/Superformula/test-full-stack/pull/7/files#diff-3f0ef24cae249263f77840910568ab23) so you can grab what you need (or all *) from just one import statement.

And I used named exports on another project for easily-importable as-needed device-sniffing properties:
```
return {
    isDesktop,
    isTablet,
    isPhone,
  }
```

fe/src/components/UserForm.js
charlieargue: Yes, absolutely! I’d even made a bookmark on July 1st to check out their Version 6 release – looks great! 

Not only the react-native support, but the performance-minded approach (relying on uncontrolled components) – not over-re-rendering and looks like built-in debounce! 

Looks like better management of error messaging and dynamic form building. Sounds like reactive forms in Angular (vs ngModel).

And looks like a great library that’s not going anywhere (lots of use, v6 just came out, active development, issues cleared up, etc…) - https://github.com/react-hook-form/react-hook-form

fe/src/components/UserForm.js
charlieargue: If I understand correctly, you’re asking about the two type=text form controls, so I’m thinking:
-	most likely abstract into a re-usable text-control, aka <TextInput /> (and create one for each type of input probably, and its surrounding/wrapping markup aka form-group)
-	as well as: include max lengths
-	switch to react-hook-forms
-	sanitize input perhaps (no special characters besides -_’ for example)
-	perhaps combine the state form variables into one state variable with { name, address, description } properties
-	make the error handling more robust and intelligent (maybe not stringify, and instead have a standardized error object returned from the server, and/or search for the actual .graphQL.errors[] objects
-	provide default values perhaps
-	localize the placeholders/error messages/etc.
-	add aria labels
-	make them responsive
-	custom validation (be able to easily apply minLength, isRequired, isEmail rules)
-	improve the UI with visual messaging/cue-ing for “touched”, “valid”, and other individual form element states (red/green borders on inputs, for example)
-	more robust form validation (preventing double-taps AKA double-form submits, on another react project, I change the submit button disabled state during form submittals, change the icon, and change the button text to “Sending…”)
With a complete form system for building dynamic forms, you might end up using something like this:
```
<TextArea name="address"
	                    placeholder={this.state.formControls.address.placeholder}
	                    value={this.state.formControls.address.value}
	                    onChange={this.changeHandler}
	                    touched={this.state.formControls.address.touched}
	                    valid={this.state.formControls.address.valid}
	          />
```
Abstracting out json objects in state that would build our forms dynamically, i.e.:
```
…
address: {
	          value: '',
	          placeholder: 'What is your address',
	          valid: false,
	          validationRules: {
	            minLength: 4,
	            isRequired: true
	          },
	          touched: false
	        },
```

fe/src/components/UserList.js
charlieargue: I assume you mean doing something like this: - https://stackoverflow.com/a/30565330/6200791

AKA creating one file with all vendor imports and perhaps another one with all my graphql imports, throw all that as non-default exports, and then be able to import all my dependencies through-out the project by only ever referencing two files! Great idea, absolutely, if I understood correctly. That along with barreling helps in big projects.

fe/src/components/UserList.js
charlieargue: Absolutely, that was on my list of todos – on another react project, using material-ui, I used Suspense and lazy along with react-router-dom to get full-page suspense loading via a full page backdrop and circular loading indicator on each page transition load, like so:
```
<Suspense fallback={
        <Backdrop className={classes.backdrop} open={true}>
          <CircularProgress color="inherit" />
        </Backdrop>
      }>
        <Switch>
          <Route path="/full-stack">
            <Hero />
            <TechStack />
          </Route>
```
I had wanted to do something similar with the react-bootstrap spinner and modal backdrops, to be triggered each API call (much like I would do in angular http-interceptors).

fe/src/fonts/OFL.txt
charlieargue: Hmm, good question -- first off, almost always, the CDN will be the fastest for font-loading, so yes.

But if you needed to / wanted to serve your own font files, and they were slow, you could start with progressive image loading (the blurry background effect) + skeletons, and in the background lazy-load a separate module with just the font files – so it doesn’t affect content-paint/loading times (just thinking out loud).

fe/src/globals.js
charlieargue: Absolutely, that should be in an .env file – I kept it out just to make your lives easier, so that when you are spinning up local .env files, you don’t have to worry about anything except the two 3rd party KEYs/URI – but this is NOT where an API base_url should go, agreed!

fe/src/queries/queries.graphql.js
charlieargue: Yes, gladly! I was using the basic esLintConfig that extends CRA, and I tried a few vscode extensions (including prettier and beautify) to be able to auto-format my gql better – but had no luck, just still new to graph QL. 

I linted both my typescript Angular front-end, and node.js typescript back-end at my previous job, and always auto-formatted.

fe/src/queries/queries.graphql.js
charlieargue: Create a type for the mutation object, and then would just pass in one object (aka “Options Object” pattern). So I guess it would be something like:
```
Mutation editUser($inputUser: UserInput!)… 
```

Then, as your input type definition grows, your mutation param code stays the same!

fe/src/ui/styles.css
charlieargue: In various .scss Sass files, or more realistically multiple Sass files and mixins, such as:
-	_variables.scss where we would define variables and their values (aka $primary color, $border-radius, etc.)
-	A main _styles.scss file perhaps that would have the main bootstrap over-rides and main site styling
-	These would come after we import the uncompiled bootstrap Sass files

And I’m still learning this approach to react component architecture, but I imagine some styles would always reside within each specific component itself (like ` import { makeStyles, useTheme } from "@material-ui/core/styles”` hooks, so ultimately you want, I imagine, all of the component-specific CSS styling to live in the component itself (as JavaScript CSS style objects, injected into style or like materia-ui does, with `className={classes.###}` 

You can then stack up app-wide styling plus component-specific styles using for eg. `className={`${classes.localClass1} `${classes.localClass2} someGlobalClass`}. Aka proper bootstrap theming.

Not to mention that material-UI has a great overrides system for changing all their internal classes, something I bet bootstrap has also, afaik – I would use that approach in production.

Just as an example, this was my `app.component.scss` file from my last big project:
```
// SING SCSS partials
@import "variables";
@import "mixins";

// bootstrap
@import "node_modules/bootstrap/scss/bootstrap"; // uncompiled SASS

// libraries
@import 'node_modules/angular2-toaster/toaster';

// font awesome, animate
$fa-font-path: "assets/fonts";
@import "~font-awesome/scss/font-awesome";
@import "~animate.css/animate";

// ngx-select styles
@import "~@ng-select/ng-select/themes/default.theme.css";

// SING overrides
@import "bootstrap-override";
@import "libs-override";

// LORIS purchasing overrides of Sing UI Template
@import "loris-purchasing-variables";
@import "loris-purchasing-colors";
@import "loris-purchasing-utilities";
@import "loris-purchasing-bootstrap";

// everything below this line is required for essential styling
@import "font";
@import "general";
@import "global-transitions";
@import "base";
@import "utils";
@import "print";


```

fe/src/components/CenteredModal.js
charlieargue: Right, so I had some questions about that. While working on the Figma files, I noticed things don’t always “seem right” between figma and the mockup? Like **font sizes,** and the **border-radius** values. 

For example, `figma::inputs.borderRadius == 8px`, but that's too much in real life imo (**I could be wrong!**), when trying to match the comps/mock-ups (I know they’re scaled down), but it looks more like **4px**?

And I’m not familiar enough with Figma yet to have gotten the margins / spacing exact. So had some questions for pixel-perfection, but decided to sacrifice that at the time and ask you guys about it here.

I did my best to match the Figma files, but sometimes things just didn’t look right with those same font-sizes, so did my best. In production, I would have stopped right away and asked for guidance and figured out why things appeared out-of-sync to me.

Like, on the Figma modal, the width is **1328px** – I succumbed to the decision to leave my width at the react-bootstrap prescribed `size=”xl”` width (around 1000px) – and since the project said NOT to do responsive, I decided that for a two-day project, approximating the Figma as-best-as-possible while making it look good was better at that point.

fe/README.md
charlieargue: No, but great idea! Then that file can be used a template showing which env variables are needed for the app – and you can base your `env.local` and `env.development|production` by copying that one!

fe/src/components/Dashboard.js
charlieargue: Just an accident from copy-paste from the react-bootstrap Modal example I used to start with, good catch!

fe/src/components/layout/ApplicationLayout.js
charlieargue: I think ` withRouter ` is unnecessary in this case/project-scope, and could be removed from this component, since it makes sense only if you need the updated match, location, and history props that withRouter gives you access to. It could be needed in the future if the project expands, and so I included.

I’m not sure if I answered your question exactly, though…

fe/src/components/UserForm.js
charlieargue: To prevent a “double tap”, I would do something similar in react-bootstrap as I just did in react material-ui recently (simplified for illustration purposes):
```
<Button disabled={isSending} type="submit">
    {!isSending && (
        <SendIcon className={classes.icon} />
    )}
    {isSending && (
        <AlarmIcon className={classes.icon} />
    )}
    <Typography
        variant="subtitle1"
        noWrap
        color="primary"
        display="inline"
    >
    {isSending ? "Sending, please wait..." : "Send Message"}
    </Typography>
</Button>
```
Not only is the button disabled while isSending (no double-tap/multi-submission possible), but the button icon and text are also changed to reflect the sending state.

charlieargue:
Hi All -

Was working on *Mateusz's* PR *questions, but they suddenly disappeared! *

Looks like he's no longer a participant on the PR?

*Here are my answers to his two questions:*

❓*Mateusz Question #1:*

*"As lodash library is quite big - do you have any ideas on how to optimize
> it (let's assume that we have our own webpack config, not CRA)?"*


In one of my last big projects (angular), I would import only the lodash
functions I needed, i.e.:
import groupBy from 'lodash/groupBy';

That reduced the size -- I could use the same approach here.

Also, I think we could use `*lodash-webpack-plugin*` to create smaller
builds.

Additionally, there’s options like tree-shaking, scope-hoisting, using
analyzers like webpack-bundle-analyzer, etc.

And of course, like lazy-loading our modules, we can code-split our bundles
up even further (i.e. `vendors.ts, polyfills.ts, main.ts, etc.` in angular)

*❓Mateusz Question #2:*

*"We query the same fields (id, name, description, address, createdAt) in
> allUsers and filteredUsers queries - do you have any idea how to unify it
> and not repeat it in every users queries? "*


Good question, good catch! Yes, sorry – those should be *fragments*!

I have updated my code and pushed it up, it’s now:
```
const fragUserResult = gql`
    fragment userResult on User {
        id
        name
        address
        description
        createdAt
    }
`

// and then used below:
// GRAPHQL queries
export const ALL_USERS = gql`
query {
    allUsers {
        ...userResult
    }
}
${fragUserResult}
`

```

Thanks guys, lots of good questions and feedback, exciting!


* Karl Golka*
*Full Stack JavaScript Developer *
contact@karlgolka.com*  |  *karlgolka.com
*cell*: (619) 581-6529

wprater:
@charlieargue I think they are all still there. You just need to expand the comments, since files changed and GitHub thinks they are irrelevant now.  :)  Thanks for your thorough responses!

charlieargue:
Sorry Will, still don't see them!

Mateusz had 2 questions, and I can't find them anywhere (expanded/loaded
more comments, but still nothing!)

I've loaded all the comments, searching for "mateusz", nothing...

Also, notice that he is no longer a participant in the code review:
[image: Screen Shot 2020-07-07 at 8.32.42 PM.png]

Thanks/sorry!

charlieargue:
Screen Shot 2020-07-07 at 8 32 42 PM

fe/src/components/layout/ApplicationLayout.js
marcin-piela-sf: Yes it's not required here and in the future will be better to use `useRouter` hook instead of HOC

fe/src/components/CenteredModal.js
marcin-piela-sf: Thanks for explanation 👍 

Those desktop designs (for resolutions bigger or equal 1366px - about 80% desktops has this resolution or bigger) are so big on purpose :)

mat-sf:
@charlieargue 
Yea I see they are no longer here, but I'm glad that you answered, good answers 👍

wprater:
Thank you.

PR #14, opened 2020-07-12
Full-Stack Code Sample

Applicant-SF:
This pull request implements the requirements in the problem statement for both the FE and BE.

See readme for information and usage.

Check out the live sample at https://applicant-sf.github.io/

Edit: There are three different committers. Those are all me - I didn't change my credentials on my different workstations.


projects/server/src/geocode.ts
```diff
@@ -0,0 +1,72 @@
...
+
+// function rawHeadersToObject(rawHeaders: string[]) : {[x: string]: string} {
+//     let headers = {};
+
+//     for(let i = 0; i < rawHeaders.length; i += 2) {
+//         headers[rawHeaders[i]] = rawHeaders[i+1];
+//     }
+
+//     return headers;
+// }
```
afaq-sf, line 17: Was this intentionally left in as commented code for reference?

projects/server/src/geocode.ts
```diff
@@ -0,0 +1,72 @@
...
+            const [, encodedAddress] = query.split('=');
+            const address = decodeURIComponent(encodedAddress);
+            console.log(address);
+
+            client.geocode({
+                params: {
+                    address: address,
+                    key: configGoogleMapsKey
+                }
+            }).then((value: GeocodeResponse) => {
```
afaq-sf, line 44: How do you evaluate when to  using `await` syntax vs promises? Do you have a preference?

projects/server/test/crud/create-request.js
```diff
@@ -0,0 +1,128 @@
+/**
+ * TODO: There is probably a library for this.
+ * 
+ * Also this is very ugly.
```
afaq-sf, line 4: In a real-world project, how do you prefer to track and resolve these sort of To Dos / technical debt over time?

projects/server/src/geocode.ts
Applicant-SF: I have removed the commented code.

Since we have to handle CORS ourselves on API Gateway, I planned to restrict the CORS header to whatever was configured. E.g. `applicant-sf.github.io`. To do that, I would have to look at the `referer` or `origin` header - depending on if I wanted to restrict the requests to a certain part of the site.

However, as you guessed, it was left in for reference. Node provides a fairly unusual data structure when dealing with raw headers odd-even pairs in a flat array. I left the function there to remind me of that. While the `headers` field has those headers typed, it doesn't have non-standard headers. I would want whatever code that handles headers to use the same strategy for standard and non-standard headers - which was why I was looking at `rawHeaders`.

Truthfully, in that function, I should have provided a reference to what I was handling anyway. https://nodejs.org/api/http.html#http_message_rawheaders

projects/server/test/crud/create-request.js
Applicant-SF: Some sort of issue tracking system. Github, Gitlab, DevOps, all do them differently with varying features. I would put this in as an issue/work-item/etc. Some even allow linking to a line in code - I know Gitlab does; Github might now-a-days.

Additonally, I wouldn't just say `this is very ugly`. I would explain why. E.g.

```
This is a spoof API Gateway request.

We should find a library or write a library that generates this spoof request. There are a lot of fixed fields here that distract from the meaningful information - see the magic strings. Those should be hidden.

Additionally, statically saved payloads may not be representative of real payloads. More research is needed.
```

projects/server/src/geocode.ts
Applicant-SF: This was strange. I typed up the answer about how `client.geocode` doesn't return a promise, it returns a 'Thenable' which Typescript complained about, but this wasn't the case. It does return a promise - so this is just a bug.

__I prefer async - await__. Callback hell was one of the early reasons I balked at Nodejs. Async-Await is being adopted in many other languages now, and people are more familiar with that paradigm.

Applicant-SF: Follow-up: I though there was an ESLint rule for `prefer-async-await` but I can't seem to find it now. My corrective action would be to enable that lint rule.

Let me do some research to see if I can find that lint rule.

projects/server/test/crud/create-request.js
Applicant-SF: As for resolving technical debt - I would consult the team. Often, I suggest devoting some percent of each sprint (if agile or agile-lookalike) to correcting tech-debt. I've seen, however, people prefer to wait for a release, and handle tech-debt next release. (Although that was an embedded-product, and the reason for waiting is a bit more compelling)

projects/server/src/geocode.ts
Applicant-SF: There is a rule, but it is a plugin. https://github.com/eslint/eslint/pull/11961.

It is not part of eslint-core.

projects/client/src/api/api-types.ts
```diff
@@ -0,0 +1,19 @@
+export interface APINextToken {
+    id: string;
+}
+
+/**
+ * TODO: Is there a clean way to share types between the server and
+ * client?
```
julian-sf, line 7: **discussion:** That's a good question.

Any ideas on how to do that?

PR #15, opened 2020-07-13
Tc sf fullstack test

TommyC5:
Hello All

Here is my pull request, hope you guys enjoy reviewing it.  

Here are where things sit in terms of the main logic of the code.

The userservice folder is pretty much the back-end, the src folder within here is where all the logic resides.  

The site folder is the front-end and the src folder should hold all the code.  

demo site

http://d20qof6e8b8d9s.cloudfront.net/

Looking forward to hearing your feedback.

Thank you.


site/src/container/HeaderBar/HeaderBar.js
```diff
@@ -0,0 +1,129 @@
...
+            let value = e.target.value;
+            props.setCurrentSearchTerm(value);
+            if (debounceFn) {
+              debounceFn.clear();
+            }
+            debounceFn = debounce(async () => {
+              return AppSyncUserServiceProvider.doSearch(value);
+            }, 1000);
+            debounceFn();
+          }}
```
julian-sf, line 103: **question:** are there any performance concerns with this code?

site/src/container/UserForm/UserForm.js
```diff
@@ -0,0 +1,258 @@
...
+      height: 300px !important;
+    }
+  }
+
+  @media (max-width: 1023px) {
+    flex-basis: 62px;
+    .google_map {
+      display: none;
+      width: 300px !important;
+      height: 300px !important;
```
julian-sf, line 67: **question:** why did you need to use `!important` here?

What are the drawbacks to `!important` and how could you work around this?

site/src/index.js
```diff
@@ -0,0 +1,23 @@
...
+import initializeGoogleMap from "./utils/initializeGoogleMap.js";
+
+initializeGoogleMap();
+
+ReactDOM.render(
+  <React.StrictMode>
+    <GlobalStyle />
+    <App />
+    <ToastContainer position="bottom-right" />
+  </React.StrictMode>,
```
julian-sf, line 19: **question:** why use `React.StrictMode`?

What benefits does it provide?  Is there a cost?

site/src/redux/Users/UsersReducer.js
```diff
@@ -0,0 +1,40 @@
...
+      };
+    case UsersActions.SET_MAX_USER_LIMIT:
+      return {
+        ...state,
+        userNumberLimit: action.userNumberLimit,
+      };
+    default:
+  }
+
+  return initialState;
```
julian-sf, line 37: **question:** what is the reason you chose to return `initialState` here instead of `state`?

Are there unintended consequences that could happen here?  Is this planned behavior?  If so, in what scenario is this useful?

userservice/src/repository/DynamoDbUserRepository.js
```diff
@@ -0,0 +1,144 @@
...
+        params.ExclusiveStartKey = data.LastEvaluatedKey;
+      }
+      data = await db.query(params).promise();
+      for (let i = 0; i < data.Items.length; i++) {
+        if (result.length >= limit) {
+          break;
+        }
+        result.push(User.fromJsonObject(data.Items[i]));
+      }
+    } while (result.length < limit && data.LastEvaluatedKey);
```
julian-sf, line 41: **question:** is it possible to go over the limit here?

If so, how would you fix this?

userservice/src/repository/DynamoDbUserRepository.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+}
+
+function randomImage() {
+  let number = Math.floor(Math.random() * 1000);
+  let result = number % 2;
+  let type = "women";
+  if (result === 1) {
+    type = "men";
+  }
```
julian-sf, line 134: **question:** would a ternary be easier here?

site/cypress/integration/Add_edit_delete_user_from_another_client.js
```diff
@@ -0,0 +1,93 @@
...
+
+describe("Updating ui base on Create, Update, and Delete from another client", () => {
+  const userStore = createStore(UsersReducer);
+  AppSyncUserServiceProvider.init(userStore);
+  before(() => {
+    cy.visit("http://localhost:3001/");
+  });
+
+  it("Can display user being added from another client", () => {
+    cy.wait(InBetweenWaitTime);
```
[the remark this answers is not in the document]
TommyC5: The wait command pretty much allows you to wait in milliseconds or wait for a resource to be resolved and move on.  You should be able to setup cypress to intercept request with cy.server and cy.route, with route having an alias.  Then use cy.wait to wait until the resource to be resolved.  When I was playing with that I was not able to get the graphql request to be intercepted and was able to get other “test” requests intercepted and came across this https://github.com/cypress-io/cypress/issues/687.    

Another possible solution I can think of is to create another test that will execute first and to ensure the page has been loaded and add a cy.get on the user card.  The whole purpose of the wait for that test was to ensure the page has been fully loaded before executing adding user from an external client.

If you know of another command in cypress, don’t mind sharing : )  As I am pretty much new to cypress.

site/src/App.js
```diff
@@ -0,0 +1,79 @@
...
+  await AppSyncUserServiceProvider.loadUsersFromState(parsed["filter"], limit);
+};
+
+window.addEventListener("hashchange", async (event) => {
+  await loadUserList();
+});
+
+initialize().then(async () => {
+  await loadUserList();
+});
```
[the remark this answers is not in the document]
TommyC5: That was my leftover experiment code in regards to playing with cypress.  I have clean that up and will be pushing the code.

site/src/container/UserForm/UserForm.js
TommyC5: I was on a mission to get the page to be responsive (the modal was the more annoying one).  So was just literally trying everything to make things look acceptable, basically taking shortcuts.  Already fix this by putting the google_map class outside the media and getting rid of the inline style that is used in map.js.

!important pretty much ignore subsequent rules.  The drawback is it can lead to confusion for other developers.  They can make a change and wonder why it’s not working.  So this is kind of like sloppy code.

However, there are business scenarios where I will use !important.  For example, I have a page/component that can be white label and embedded, I need to expose a certain set of CSS class to my clients and allow them to change certain look and feel.   I don’t ever want them to touch the dimensions like height/width and so on, that is when I will set !important to them.

userservice/src/repository/DynamoDbUserRepository.js
TommyC5: Yes, already made that change.  Thank you !

userservice/src/repository/DynamoDbUserRepository.js
TommyC5: I don’t believe so, I wrote a unit test and an integration test (testing on a local Dynamodb) that generates 100 different test with 2000 users with a random number limit that goes up to 2000 and all those 100 test passed.  Should have the code push with those tests.  I could be missing something here entirely, especially anything Dynamodb related, pretty much been only playing with that for a week now. So feel free to point out the mistake I made here.

site/src/components/UserCard/UserCard.js
```diff
@@ -0,0 +1,120 @@
...
+    float: right;
+    font-size: 16px;
+    padding-top: 5px;
+    p {
+      display: inline;
+    }
+    label {
+      margin-left: 5px;
+      color: #b42320;
+    }
```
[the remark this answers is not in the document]
TommyC5: Those classes don’t need to be there, at first I was trying to come up with my own design.  Color scheme and all heh, so I was pretty much prototyping.  Then I post the question can I come up with my own design and was told to stick with the Figma design heh.  So went back and try to make this pixel-perfect and those were leftovers.    

To do this without classes if you want to target another style component you can just put ${ComponentName}.  My recommendation is to stick with the $(ComponentName} approach if is your own code.  If you are using third party components (in this homework assignment, that google map control I used is a great example) and you need to change its look and feel I think the only way via style component is to add the class name of the third party component you set for.

site/src/components/Map/Map.js
```diff
@@ -0,0 +1,69 @@
...
+
+let markers = [];
+
+const Map = (props) => {
+  const [map, setMap] = React.useState(null);
+  const onLoad = React.useCallback((map) => {
+    const bounds = new window.google.maps.LatLngBounds();
+    map.fitBounds(bounds);
+    setMap(map);
+  }, []);
```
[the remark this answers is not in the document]
TommyC5: The useCallBack will make a memorize function.  Every time a react component render it will recreate these function.  So the useCallBack second parameter is an array and you can pass in a different value for it to determine if the function should be recreated.  As for passing in an empty array that is like declaring a function outside of that react component, so it will be created one time.  In this case, I needed to get back a reference of the actual google map to plot marker against.  So that was I I use the useCallback hook.

site/src/index.js
TommyC5: StrictMode helps to identify with problems with your react app.  There is a performance cost, but this should not affect production build cause it only runs in development mode.

site/src/redux/Users/UsersReducer.js
TommyC5: This is a mistake on my part, it supposed to be state.  The consequences are pretty much if an unknown action gets passed in it can put the app into a bad state.  I.e if that reducer ever get use in a combined reducer with that code.

site/src/components/Map/Map.js
```diff
@@ -0,0 +1,69 @@
...
+          );
+        }
+      });
+    } else {
+      map.setCenter({
+        lat: 0,
+        lng: 0,
+      });
+    }
+  };
```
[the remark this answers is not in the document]
TommyC5: No this is an oversight on my part completely,  I was pretty much getting back into the React world so to speak and learning old and new things on the fly.  I pretty much mess this part up.  I clean up the code and should be better now.  Thank you for pointing this out.

site/src/components/Map/Map.js
```diff
@@ -0,0 +1,69 @@
...
+  const [map, setMap] = React.useState(null);
+  const onLoad = React.useCallback((map) => {
+    const bounds = new window.google.maps.LatLngBounds();
+    map.fitBounds(bounds);
+    setMap(map);
+  }, []);
+
+  const onUnmount = React.useCallback((map) => {
+    setMap(null);
+  }, []);
```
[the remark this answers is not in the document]
TommyC5: I believe I have answered this question with the two answers I have below.

site/src/container/HeaderBar/HeaderBar.js
TommyC5: I am just going to be guessing here, might be the async I put in there?  Don’t think it has to do with SyntheticEvent since I am not referencing it later.  Please feel free to point out my mistake.  Thank you.

site/cypress/integration/Add_edit_delete_user_from_another_client.js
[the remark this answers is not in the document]
julian-sf: > If you know of another command in cypress, don’t mind sharing : )

Sure!  Generally [`.wait`'s are considered an anti-pattern in Cypress](https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting).  The general approach is to use `cy.get` on something that should be visible only when the page is loaded.  In this case, I would assume you could wait for the text that shows an empty user list, since you won't load the user list until after this even occurs.

site/src/container/HeaderBar/HeaderBar.js
julian-sf: Let me rephrase the question :D

How would you keep this function from being redeclared every render cycle?  Does it matter?

userservice/src/repository/DynamoDbUserRepository.js
julian-sf: No mistake, just wanting to see your reasoning, haha!

site/src/container/HeaderBar/HeaderBar.js
TommyC5: You have given me so much hint in your past questions and I completely glance over this (looking embarrassed here heh).  Basically “useCallback” to create a memorize function.  So based on all these hints you given me I went back and did some optimization.  Pretty much made sure those UserCards don’t get re-render, the search part,  modal, and map by using memo.    

So is there any more hint you can give me?  Like anything else that catches your eye that is worth mentioning heh?

site/cypress/integration/Add_edit_delete_user_from_another_client.js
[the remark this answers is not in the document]
TommyC5: I guess my idea of having a test that runs first to make sure the list of user cards with cy.get was pretty close to yours .  But I guess another test is not needed and this would be a better approach.  Thanks for the tip, appreciated on my end.

wprater:
Thank you for your submission.

PR #19, opened 2020-07-30
[wip] Fullstack

vpcsilva:
Hi Guys,

This is my PR request for this code test.

The backend api is deployed to aws. The url to the api is https://df6bsei4ovh57mw66gqlnjw73q.appsync-api.us-east-1.amazonaws.com/graphql .

Unfortunately I wasn't able to make the front-end to work on time so I pushed my changes and created this wip PR.

I'll be happy to answer any questions.


vpcsilva:
The api key is: `da2-fibvzizvbjghjfr545nsdqfkii`.

wprater:
Hi.  Is this PR a WIP or ready for review?  The title still has a `WIP` in it, so I wanted to be certain.  Also, the preview URL is not accessible. 

```
{
  "errors" : [ {
    "errorType" : "UnauthorizedException",
    "message" : "You are not authorized to make this call."
  } ]
}```

**Comment by vpcsilva** [comment]
Yes, it's a WIP because I want to send the front end code today, if its possible.

**Comment by mat-sf** [comment]
@vpcsilva please send it and remove `[WIP]` from the title when ready

File: api/schema.graphql [hunk 1]
```diff
@@ -0,0 +1,52 @@
+schema {
+  mutation: Mutation
+  query: Query
+}
+
+type Mutation {
+  createUser(
```
marcin-piela-sf, line 7: What are your thoughts on making input as 
```
input CreateUserInput {
 name: String!
 dob: String
 address: String
 description: String
}

createUser(input: CreateUserInput)
```

api/schema.graphql
```diff
@@ -0,0 +1,52 @@
...
+  type: String
+  coordinates: [Float]
+}
+
+type Location {
+  geometry: Geometry
+}
+
+type UserList {
+  items: [User]
```
marcin-piela-sf, line 40: Is it possible that `items` will be `null` or `undefined` here?

api/schema.graphql
```diff
@@ -0,0 +1,52 @@
...
+    dob: String
+    address: String
+    description: String
+  ): User
+}
+
+type Query {
+  user(id: String!): User
+  users(limit: Int, nextToken: String): UserList
+  filterUsers(name: String, limit: Int, nextToken: String): UserList
```
marcin-piela-sf, line 26: Can you help me understand the value of this query being separate from users? What are your thoughts on removing it and using `users` instead? Would you recommend it?

api/schema.graphql
```diff
@@ -0,0 +1,52 @@
...
+  getLocation(address: String!): Location
+}
+
+type Geometry {
+  type: String
+  coordinates: [Float]
```
marcin-piela-sf, line 32: Is it possible that `coordinates` will be null or undefined here?

.gitignore
```diff
@@ -0,0 +1,119 @@
...
+
+# Next.js build output
+.next
+out
+
+# Nuxt.js build / generate output
+.nuxt
+dist
+
+# Gatsby files
```
marcin-piela-sf, line 87: Would you recommend using such `.gitignore` in real projects? I mean `.gitignore` which handles apps like Gatsby, Nuxt etc which are not used in this project.

api/handler.ts
```diff
@@ -0,0 +1,15 @@
+import { Logger } from './src/utils';
```
marcin-piela-sf, line 1: In a real-world implementation, do you recommend creating directories like `utils`?

api/mapping-templates/updateUser.request.vtl
```diff
@@ -0,0 +1,48 @@
+{
```
marcin-piela-sf, line 1: What are advantages and disadvantages of creating .vtl files instead of eg. creating lambda to handle domain logic?

api/serverless.yml
```diff
@@ -0,0 +1,200 @@
...
+        request: 'updateUser.request.vtl'
+        response: 'updateUser.response.vtl'
+      - dataSource: LambdaLocation
+        type: Query
+        field: getLocation
+        request: 'getLocation.request.vtl'
+        response: 'getLocation.response.vtl'
+
+    dataSources:
+      - type: AMAZON_DYNAMODB
```
marcin-piela-sf, line 83: How would you recommend enable autoscaling for this table?

api/src/services/mapBox/index.ts
```diff
@@ -0,0 +1,52 @@
+import axios from 'axios';
```
marcin-piela-sf, line 1: What are benefits of using `axios` over eg. `unfetch`?

client/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "name": "client",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^4.2.4",
+    "@testing-library/react": "^9.5.0",
+    "@testing-library/user-event": "^7.2.1",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "react-scripts": "3.4.1"
```
marcin-piela-sf, line 11: Do you recommend using create-react-app in real production apps?

marcin-piela-sf (commented):
Hi Victor, thanks for the PR. I left some comments and looking forward to your answers/thoughts 👍

api/handler.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getLocation } from './src/controllers/location';
+
+export const locationHandler = async (event: any) => {
+  const log = new Logger();
+  log.info({ event }, 'getLocation Handler!');
+
+  try {
+    const { address } = event;
+    return getLocation(address, log);
```
marcelino-sf, line 10: How would you recommend catching also errors in an asynchronous way?

api/mapping-templates/filterUsers.request.vtl
```diff
@@ -0,0 +1,15 @@
+{
+    "version" : "2017-02-28",
+    "operation" : "Scan",
```
marcelino-sf, line 3: Could you explain to me when to use `Scan` over `Query` and otherwise? What would be the problems of using it in a real-world project?

api/serverless.yml
```diff
@@ -0,0 +1,200 @@
+service:
```
marcelino-sf, line 1: What are your thoughts on using Serverless Framework over Terraform?

api/src/config/default.ts
```diff
@@ -0,0 +1,14 @@
+export default {
```
marcelino-sf, line 1: Can you explain the advantages of exporting a default literal object over exporting constants?

api/src/config/index.ts
```diff
@@ -0,0 +1,8 @@
+import { merge } from 'lodash';
```
marcelino-sf, line 1: Could you explain the advantages of using `import { merge } from 'lodash'` over `import merge from 'lodash/merge'`?

api/src/errors/badRequest.exception.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * NotFoundException class.
+ */
+export default class BadRequestException extends BaseException {
+  /**
+   *
+   * @param {string} msg
+   * @param {any} data
+   */
```
marcelino-sf, line 11: Can you explain the advantages of having empty jsdoc over not having it?

api/src/errors/forbidden.exception.ts
```diff
@@ -0,0 +1,15 @@
+import BadRequestException from './badRequest.exception';
+
+/**
+ * NotFoundException class.
+ */
+export default class NotFoundException extends BadRequestException {
```
marcelino-sf, line 6: Do you think named exports would prevent such errors?

api/src/utils/logger.ts
```diff
@@ -0,0 +1,31 @@
+import * as Pino from 'pino';
+import Config from '../config';
+import { ILoggerConfig } from '../interfaces';
+import { merge } from 'lodash';
+
+export class Logger {
```
marcelino-sf, line 6: Could you explain the importance of having such abstractions?

api/src/errors/badRequest.exception.ts
```diff
@@ -0,0 +1,15 @@
+import BaseException from './base.exception';
+
+/**
+ * NotFoundException class.
+ */
+export default class BadRequestException extends BaseException {
```
marcelino-sf, line 6: Can you help me understand the importance of code comments in general? What are the best practices in your opinion?

marcelino-sf (commented):
Hey @vpcsilva, thank you for your time creating the pull request, I left a few questions and I'd be happy to see your thoughts.

api/schema.graphql
vpcsilva: Using inputs is very useful when you have multiple mutations using the same parameters schema. In my code, for example, I could have used this `CreateUserInput` in the `createdUser` and `updateUser`mutations, making the schema file cleaner and easier to maintain.

api/schema.graphql
vpcsilva: Yeah, the `items` property can be null there, because I didn't use a non-nullable modifier, although I'm not sure if AppSync would ever return null in this case. But using something like `items: [User!]!` would garantee a more reliable response, since it would only allow the return to be an empty list of a list with not null values.

api/schema.graphql
vpcsilva: To be honest I didn't thing much about this at the time. Looking back at this today, it makes more sense to have the search parameters within the `users` query. Maybe I did it like this because of the AppSync mappings. I would recommend having a  search parameter in the `users` query, with the search parameters (in this case, the name).

api/schema.graphql
vpcsilva: It's the same situation as the `items` field above. As it is, it'l allowing `coordinates` to be `null` or `undefined`, or to return a list with `null` items.

.gitignore
vpcsilva: I used the gitignore package to create this file (`npx gitignore node`), without taking time to clean it up. So no, I wouldn't leave it like this in a real project.

api/handler.ts
vpcsilva: I'm not against having folders like `utils/` or `shared/` as a place for reusable and shared functions present in the codebase. But in this case, I think the `Logger` is more like a module in the project, with it's own configurations and logic, than just an utility function.

api/mapping-templates/updateUser.request.vtl
vpcsilva: It was interesting to use `.vtl` files in this project, to see how it works. In my opinion:

**Advantages**:
* On simple queries and mutations where you don't have any complicated business logic, it is simple and quick to just write a small `.vtl` file and let AppSync do the rest.
* Having everything managed, including things like communication with database, means less pieces to maintain.

**Disadvantages**:
* The learning curve, taking in account every dev in the team. It is not much, but it's something to take in consideration.
* The application is completely coupled with AWS.
* Anything more complex than a very simple read, write or search, becomes maybe too complex in the `.vtl` files and might not be worth the trouble of going for it in the first place. Take for example the `updateUser.request.vtl` file. To maintain it's logic can be a problem.

api/serverless.yml
vpcsilva: In the serverless file you can either configure the Scaling Policy, in the resources property, or, in a simpler and quicker way, we can use the `serverless-dynamodb-autoscaling` plugin to easily configure scaling parameters, without having to manually do it in a more verbose fashion. Personally, I'd go for the second option, unless it is really necessary to do it manually.

api/src/services/mapBox/index.ts
vpcsilva: Axios is a pretty common library for making requests in javascript. It's top of mind and well known among javascript programmers, so everyone in the team would have seen it before and might know how it works. That's why I used it here, because it's just at the top of my mind and I was not worried about the size or performance of the libbrary in this application.

client/package.json
vpcsilva: I believe create-react-app is great to quick start an application and start developing right away. But in producution you might want to have more control over the bundler configurations and optimizations, so detatching from CRA or write the bundling configurations from scratch would give you more control over those things.

api/handler.ts
vpcsilva: Here I could have usued await for the result of the function, like 
```javascript
const result = await getLocation(...);
return result;
```
or wrapped it in a promise doing something like:
```javascript
return getLocation(...)
.catch(error => {
  log.error(...)
  throw error
});
```

api/src/config/default.ts
vpcsilva: Here I don't see any advantages of one over exporting constants. I think the standard may vary from team to team.

api/src/config/index.ts
vpcsilva: The only advantage in this case is that the first is "_easier_" to write than the second. But the second allows you to import only the `merge` function code, which gives you a smaller import.

api/src/errors/badRequest.exception.ts
vpcsilva: I don't see an advantage here. I'm not sure if I were going to write a description for this constructor (which I don't think so), or I generated this with a shortcut and forgot about it.

api/src/errors/forbidden.exception.ts
vpcsilva: Yes, it would definitely help to spot and fix errors like this, since I would not be able to import this class with a different name if I were using a named export.

api/src/utils/logger.ts
vpcsilva: Here I believe I didn't need for such abstraction, I could export just the instance of Pino with the configurations loaded. Maybe if the project needs something specific regarding the logging process, it would be necessary. But in the majority of cases, it just isn't.

api/src/errors/badRequest.exception.ts
vpcsilva: I used to believe that if the code is well written enough, comments wouldn't be necessary, but I've seen a lot of cases where comments helps you to quickly understand a piece of code or a function, allowing any member of the team to quickly start working on maintaing someone else's code.

I don't like to fill up the code with unnecessary comments on things that are self explanatory, like this example. But having jsdocs on a more complex business function would allow anyone to quickly know what is the goal for that function without having to read and understand the code.

One other thing that I like to do is to use line comments on logical steps of a complicated function, to help the person who is reading to understand what each piece of code on that function is going to do.

vpcsilva:
Thanks @marcin-piela-sf and @marcelino-sf for taking the time to review the code, I really appreciate it. Let me know if you need anything else from me.

api/serverless.yml
vpcsilva: I think I skipped this one.
Serverless is great to get things up and running quickly, with a vast collection of plugins that helps us to easily configure the app infrastructure. But it's not the best to configure resources from other services beyond the _serverless_ ecosystem. In a bigger application, using multiple cloud services, like sqs, ses, sns, cognito, and others, Terraform is much more powerful and capable to deal with those situations, besides the possibilities of having custom reusable modules within a company, maintaining a standardized use of these resources between the teams.

wprater:
Thank you @vpcsilva !

PR #22, opened 2020-08-06
Brian S

bsantare-sf:
Hello all,

Please find the details for this submission in the README.md.

I'm looking forward to your feedback!

Thanks!

Brian Santarelli


marcin-piela-sf:
@bsantare-sf thanks for the PR :) Could you push `yarn.lock` file as well, I want to be 100% sure that I see the same version locally as you :)

bsantare-sf:
Sorry about that! yarn.lock is now in the repo

package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "lint-staged": {
+    "src/**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
+      "eslint . --fix --ext js,jsx,ts,tsx"
+    ]
+  },
+  "devDependencies": {
+    "@commitlint/config-conventional": "^9.1.1",
+    "commitlint": "^9.1.0",
+    "electron": "^9.1.2",
```
marcin-piela-sf, line 22: Why "electron" is in devDependencies? I don't see any usage of it

user-mgmt-app/public/index.css
```diff
@@ -0,0 +1,7 @@
+body {
+  width: 1800px;
```
marcin-piela-sf, line 2: Could you elaborate more why body has fixed width and height?

user-mgmt-app/src/App.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <ThemeProvider theme={DefaultTheme}>
+        <HashRouter hashType="slash">
+          <Routes />
+        </HashRouter>
+      </ThemeProvider>
+    </>
+  );
+}
+
+export default withErrorBoundary(withApolloClient(App));
```
marcin-piela-sf, line 23: Could you elaborate more on why HOCs instead of simple functional component attached inside App tree?

user-mgmt-app/src/components/avatar/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { AvatarDiv } from 'components/avatar/styles';
+
+export interface AvatarProps {
```
marcin-piela-sf, line 4: is it necessary to export this interface? And could you elaborate more on why `interface` instead of `type`?

user-mgmt-app/src/components/avatar/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+  imageUrl: string;
+  className?: string;
+}
+
+const AvatarComponent: React.FC<AvatarProps> = ({ imageUrl, className }) => {
```
marcin-piela-sf, line 9: Could you elaborate more on why did you use `React.FC`?

user-mgmt-app/src/components/edit-user/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { User } from 'graphql/user-api/@types';
+import { useUpdateUserMutation } from 'graphql/user-api/mutations/generated/userapi-mutations';
+import { useDebounce } from 'hooks/use-debounce';
+import { UseGeocodeResult, useGeolocation } from 'hooks/use-geolocation';
+import { InteractiveMap, Marker } from 'react-map-gl';
+import React, { useEffect, useState } from 'react';
+import { Optional } from 'types';
+
+// Mapbox token scoped to read map access
+// TODO: Secure this
```
marcin-piela-sf, line 27: How would you secure this token?

user-mgmt-app/src/components/edit-user/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+      }
+    } catch (e) {
+      console.error(toErrorStr(e));
+      setErrorMessage('Error updating user');
+    } finally {
+      setSaving(false);
+    }
+  };
+
+  const handleNameChange = (newName: Optional<string>) => {
```
marcin-piela-sf, line 85: What form library would you choose in react project to avoid writing such functions?

user-mgmt-app/src/components/edit-user/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  <InputContainer>
+                    <Input
+                      name="name"
+                      label="Name"
+                      value={newInputState.name}
+                      placeholder="Name..."
+                      onChange={(newValue: Optional<string>) =>
+                        handleNameChange(newValue)
+                      }
+                      width="611.5px"
```
marcin-piela-sf, line 201: Do you prefer to write fixed widths outside styles (css or styled-components)?

user-mgmt-app/src/components/input/styles.tsx
```diff
@@ -0,0 +1,43 @@
...
+`;
+
+export const StyledInput = styled.input<{
+  width?: string;
+  height?: string;
+}>`
+  width: ${(props) => props.width || '400px'};
+  height: ${(props) => props.height || '64px'};
+  background: #ffffff;
+  border: 1px solid rgba(0, 0, 0, 0.1);
```
marcin-piela-sf, line 20: How would you share such values across the project?

user-mgmt-app/src/components/user-list/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { color } from 'theme/token';
+
+export interface UserListProps {
+  users: User[];
+  onEdit: (user: User) => void;
+}
+
+const UserListComponent: React.FC<UserListProps> = ({ users, onEdit }) => {
+  const userCards = users.map((user, index) => (
+    <UserCardContainer>
```
marcin-piela-sf, line 17: What is missing here? 🤓

user-mgmt-app/src/hooks/use-user-data.tsx
```diff
@@ -0,0 +1,284 @@
...
+    // eslint-disable-next-line
+  }, [pagesToLoad, cursorState.isMoreData]);
+
+  /**
+   * Update to a new filter value.  This function will first attempt to filter loaded data and if no data is found,
+   * make a page load call.
+   *
+   * @param userNameFragment The new user name filter value
+   */
+  const updateFilter = useCallback(
```
marcin-piela-sf, line 216: Can you elaborate more on why you used useCallback here? Do you have any specific pattern when to use useCallback/when not to use?

user-mgmt-app/src/theme/default-theme/global-styles.ts
```diff
@@ -0,0 +1,81 @@
...
+      url('fonts/source-sans-pro-v13-latin-600.woff2') format('woff2'),
+      /* Chrome 26+, Opera 23+, Firefox 39+ */
+        url('fonts/source-sans-pro-v13-latin-600.woff') format('woff'); /* Chrome 6+, Firefox 3.6+, IE 9+, Safari 5.1+ */
+  }
+
+  .nameLabel {
+    font-family: Source Sans Pro, sans-serif;
+    font-style: normal;
+    font-weight: 600;
+    font-size: 21px;
```
marcin-piela-sf, line 33: Can you elaborate more on why you decided to use px unit for font-size instead of eg. rem?

user-mgmt-app/src/components/loading/styles.tsx
```diff
@@ -0,0 +1,44 @@
...
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  align-content: center;
+  padding: 15px;
+  background: #f8f8f8;
+  border-radius: 8px;
+
+  font-family: Source Sans Pro, sans-serif;
```
marcin-piela-sf, line 25: Is it necessary? Font-family is defined in global style

README.md
```diff
@@ -1,112 +1,182 @@
...
+  * `AWS API Gateway/Lambda/DynamoDB/CloudFormation`
+  * `TypeGraphQL` - GraphQL server implementation via Typescript decorators
+  * `typedi` - tiny Typescript IoC container for dependency injection
+  * `winston` - logging
+* Front end
+  * `React`
+  * `Apollo client`
+  * `Styled System`
+  * `Storybook`
+  * `Cypress`
```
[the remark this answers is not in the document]
bsantare-sf: I planned on writing a simple Cypress E2E test but didn't get to it.  I have not used Cypress before but have used Selenium with Cucumber in the past.

package.json
bsantare-sf: Huh, honestly I have no idea how that got there.  That was definitely an accident/oversight.

user-mgmt-app/public/index.css
bsantare-sf: I meant to remove this.  I put this in place to aid in getting the initial desktop layout pixel perfect and forgot to remove :(

user-mgmt-app/src/App.tsx
bsantare-sf: The HOC gives the flexibility to wrap components in the tree without modifying the wrapped component code and thus aids in encapsulation.  i.e. in the case of withApolloClient, if the ApolloClient initialization logic were to change, the HOC would encapsulate that change rather than needing to touch code wherever ApolloProvider is wrapping a component tree to update the init logic. 

As a more concrete example, we are currently integrating a new UI component library into a project which requires different theming support than the rest of the app "page" level components.  We wrote a simple HOC that wraps a component with a theme provider and another required component.  This enables developers to simply wrap new page components that use this new library with the HOC and if anything in the HOC changes (how themes are managed etc.), the page level component doesn't need to change.  When the full application is migrated, the HOC can be moved up to the root element and removed from the page components in a simple refactor without any other component modifications.

user-mgmt-app/src/components/avatar/index.tsx
bsantare-sf: In the scope of this application, probably not.  It's only really useful for extension.  I am honestly just in the habit of building components this way and a private type would work just as well in this case.  An example of an extension would be to have a new component SuperSpecialAvatar: React.FC<SuperSpecialAvatarProps> where SuperSpecialAvatarProps extends AvatarProps and adds some properties.

This question prompted me to do some reading on the topic :). It seems that most favor using type to describe the shape for FCs.  Also, I learned that an interface can extend a type (but not vice-versa) so there is really no argument for using an interface over a type for component props.

user-mgmt-app/src/components/input/styles.tsx
bsantare-sf: A couple of ways:
1.  A global css class (either in a style sheet or css in js global)
2. I have recently done some work with Styled System.  It's pretty easy to create custom variants that can be overlaid onto a component via css in js.  For example, we recently implemented a text variant for a theme that encapsulates lineHeight, fontFamily, fontWeight and fontSize by a design token - i.e. "title" so this can be used across components by setting textStyle="title" on the component and the themed group of style props is applied to that component.

user-mgmt-app/src/components/user-list/index.tsx
bsantare-sf: 😞   I'm not sure

user-mgmt-app/src/hooks/use-user-data.tsx
bsantare-sf: UseCallback will only re-render a function when one of its declared dependencies change.  Generally, I use useCallback whenever a function is not trivial and re-rendering will impose more cost than the cost of useCallback itself.  This is generally any function that isn't a simple inline.

user-mgmt-app/src/theme/default-theme/global-styles.ts
bsantare-sf: Honestly, that value came straight from Figma - there is definitely no good reason it can't be rem.  I'm not super strong with css but do understand that rem is viewport responsive.  
Another approach would be to use responsive font sizes with breakpoints whether they are px or rem (i.e. Styled System).

user-mgmt-app/src/components/loading/styles.tsx
bsantare-sf: Nope - not necessary.

user-mgmt-app/src/components/avatar/index.tsx
bsantare-sf: Sorry - I missed these couple of comments as they were collapsed  :( 

I used React.FC as this component is stateless - i.e its a purely functional component.  All of the data the component uses is passed into the component function via props.

user-mgmt-app/src/components/edit-user/index.tsx
bsantare-sf: In this specific instance (MapBox), I would:
1) make sure the token scopes are the minimum required for the use case of the token 
2) add URL restrictions to the token in the MapBox control panel so it is only usable when requests originate from the application origin
3) I would remove the token from source control and inject it into the serving environment and render it into the page when served.

user-mgmt-app/src/components/edit-user/index.tsx
bsantare-sf: I have used Formik in the past but would likely consider React Hook Form or something similar as well.

user-mgmt-app/src/components/edit-user/index.tsx
bsantare-sf: Generally no.  Most React applications I have worked on have been designed to be responsive with pre-defined breakpoints via media queries for containers, elements, padding, margins etc.

lambda-api/serverless.yml
```diff
@@ -0,0 +1,117 @@
...
+                KeyType: RANGE
+            Projection:
+              ProjectionType: ALL
+            ProvisionedThroughput:
+              ReadCapacityUnits: 1
+              WriteCapacityUnits: 1
+
+functions:
+  lambdaapi:
+    handler: src/handler.gqlRequestHandler
```
afaq-sf, line 107: I noticed you're using API Gateway + Lambda + apollo-server-lambda.

Can you help me understand the costs/benefit for alternative approaches? For ex: managed GQL service like AppSync or hosted Apollo server instance

lambda-api/serverless.yml
```diff
@@ -0,0 +1,117 @@
...
+    User: 'User'
+  indexNames:
+    User_Name: user-global-si-name
+  endpoints:
+    dynamodb-url: 'http://localhost:8000'
+
+resources:
+  Resources:
+    User:
+      Type: AWS::DynamoDB::Table
```
afaq-sf, line 76: How would you recommend enable autoscaling for this table?

README.md
```diff
@@ -1,112 +1,182 @@
...
+1) A User CRUD resolver
+2) A geocoding resolver (backed by the Mapbox SaaS)
+
+#### Logging strategy
+Winston logging was integrated into the Lambda stack.  Winston is configured to log to CloudWatch when deployed in an AWS environment
+
+#### Retry strategy
+Currently the Serverless framework does not provide a way to configure retries at the gateway level.  Even though retries are not configurable, it is still possible to configure a dead letter queue (DLQ) via the `onError` property for a function in the serverless.yml configuration file.  To enable a DLQ, simply provide the AWS `arn` for the SNS topic of interest.
+
+Retries are also implemented in the SPA via a retry middleware configured for the Apollo client.  This policy is configure for *only* retrying on a failed query attempt and not for mutations
```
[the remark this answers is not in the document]
bsantare-sf: A couple of thoughts here:

1.  The SPA side should use retries with exponential back-off and a randomized bias so as to prevent dogpiling during short-duration outages (i.e. ephemeral network outages)

2.  With a normal microservice, this type of problem is handled by a circuit breaker either at the API level and/or at the downstream service invocation level that either fails fast when the endpoint is out of commission, returns a previously cached result or returns a default response.  This same pattern can be applied for downstream service calls _within_ the lambda with the caveat that circuit breaker state would need to be retrieved/updated from an external service (i.e. Redis) which could also lead to downstream failures :(  As far as handling failures at the resolver level, something would need to be inserted between the API gateway and the resolver lambda (i.e. another lambda) that acts as a circuit breaker for the resolver.  The same limitations apply here as the state of the resolver circuit breaker lambda would need to be stored in an external store as well.

README.md
```diff
@@ -1,112 +1,182 @@
...
+### Composition/topology
+The back end consists of an Amazon AWS API Gateway endpoint, an AWS Lambda function and an AWS DynamoDB data store.  Incoming GraphQL POST requests are handled by the API gateway and are forwarded on to the GraphQL Lambda handler to be fulfilled.  
+
+#### Lambda function composition
+The Lambda function is composed of two GraphQL resolvers
+1) A User CRUD resolver
+2) A geocoding resolver (backed by the Mapbox SaaS)
+
+#### Logging strategy
+Winston logging was integrated into the Lambda stack.  Winston is configured to log to CloudWatch when deployed in an AWS environment
```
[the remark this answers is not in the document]
bsantare-sf: I would generally have some logging abstraction such as Winston in place within the application code no matter what logging infrastructure is in use.  This enables categorized logging (i.e. debug, info, error) at the application code level that can be switched at runtime for debugging, etc.  The back end to this logging abstraction will push log messages in the appropriate format to the appropriate place whether it is CloudWatch or console where the consuming application(s) can use them.  Application metrics (internal invocation timing) can also be captured and published as well although this may differ depending on the collection services in use.  i.e. some monitoring infrastructure extract application metrics from custom log message formats while others require your application to push metrics via API.

I have used the ELK stack and DataDog on previous projects both of which make log searching and application metrics display pretty simple and approachable.  Cloudwatch has similar capabilities but some of these third party tools are more approachable in my opinion.

lambda-api/serverless.yml
bsantare-sf: AppSync:
Advantages:
1. Provides a more integrated way to create GraphQL-centric infrastructure via a schema-first approach which will lower implementation costs.  i.e. simple Cognito integration etc.
2. The back end for resolvers can be almost anything including lambda, REST, other AppSync instances, etc.  It even supports direct calls to some AWS services - i.e. Dynamo.  I honestly think coupling in this way (the Dynamo example) is a bad thing though :/
3. Provides support for GraphQL subscriptions out of the box with offline restoration capability.  i.e. an offline client application can efficiently be brought up to current without restoring its entire state 
Disadvantages:
1. Vendor lock-in - using AppSync on the client-side locks your application into AppSync on the back end
2. Centralized schema management - see Apollo Federation below

Hosted Apollo Server:
Advantages:
1.  Apollo Federation - Services can register themselves with the Apollo Server infrastructure and their schema will be "stitched" into the exposed GraphQL schema for the federation.  This enables the independent evolution of schema for bounded contexts without centralized management.  For instance, to add a query to service, add the new GQL query, test, deploy and it will be available on the stitched GQL schema once re-registered at deployment time
Disadvantages:
1.  No subscription support as of now
2. Much more complex to integrate with other AWS services

lambda-api/serverless.yml
bsantare-sf: Since this is a Serverless-managed lambda, I would utilize the _serverless-plugin-dynamodb-autoscaling_ plugin to apply autoscaling settings during deployment.  This plugin enables global (all Serverless managed tables) or customized (per table) configuration of the following CloudFormation autoscaling parameters:
ScalableTarget.MaxCapacity
ScalableTarget.MinCapacity
ScalingPolicy.TargetTrackingScalingPolicyConfiguration.TargetValue
ScalingPolicy.TargetTrackingScalingPolicyConfiguration.ScaleInCooldown

As far as settings for these values, I would start with the default values and monitor Cloudwatch metrics and determine appropriate steady-state, ceiling, and cool-down values.  This would be an iterative process that would require monitoring as load changes over time.

PR #24, opened 2020-12-01
Coding Challenge Solution

gztomas:
This is the proposed solution for a demo app that displays and provides edition for a user list.

Please try an online demo of this code [here](https://dev.d2fe8k909cfdzv.amplifyapp.com)

Please check the README for instructions on how to setup, run, test.

### AppSync for gql API + DynamoDB for data source

This was the recommended approach and AppSync already provides queries and mutations for the CRUD operations we need here. A DynamoDB data source, and elastic search. Given authentication was not a requirement for this project, I focused on getting the other requirements right, we are using just an API key that is publicly available.

### Apollo Client for GQL client and data caching

Besides using Amplify, I find their GQL client to be quite limited, as it does not provide server state management. Instead, preferring the Apollo client here. Its normalized cache gives us updates to the user lists for free from the updateUser mutation. Just added a field policy (see `src/apolloClient.ts`) for managing the simple pagination requirements we have here.

### Next.js

Not using SSR for this project, but leveraging SSG, removing the requirement for configuring a bundler.

### Jest + Playwright

Both unit and functional tests are run by jest. User list with search and edit modal features are covered but functional tests powered by Playwright. That includes both functional and visual regression tests. Functional tests are run against a mocked backed using msw.
Two things I value about this setup

- If ran with iTerm, failing functional tests will output an inline screenshot.

- A `jest-screenshot-report` folder will be generated with an image compare tool in case any visual regression test fails.

Due to time constraints, coverage could be improved. Existing tests work as a proof of concept. `src/components/Mapbox.test.tsx` is an example for unit tests, while `src/__functional__/users-list.functional.ts` is for functional tests.

### User Interface

This application is built on top of the building blocks provided in `components/system`. Those are styled-components, a theme, some global styles, and a couple of hooks.

#### A note about the UI mock

I found the dimensions in the provided mock to be a little big for many screens. To deliver a good experience I decided to base all dimensions (except borders and shadows) in rem units and use a media query to adjust root font size. For screens width 1440px or lower the font-size is reduced to 12px. For bigger screens, we keep the dimensions provided by the mock. Please check `src/components/system/theme.ts` to see the implementation for this.

## Wishlist

Things I would like to add with more time

- Focus trap for the modal
- Server-side rendering
- Better tests coverage


src/components/UsersList.tsx
```diff
@@ -0,0 +1,133 @@
...
+          value={searchTerm}
+          onChange={(e) => setSearchTerm(e.target.value)}
+          data-testid="users-list-search-input"
+        />
+      </Header>
+      <Grid>
+        {items.map((item, i) => (
+          <UserCard
+            key={item?.id ?? i}
+            onClick={() => setSelected(item)}
```
alex-sf, line 61: Not a big deal here but inlining anonymous functions break shallow prop equality checks. Doesn't really have any meaning performance impact on this application but could cause issues if you are passing them into more expensive components.

src/components/UserEditor.tsx
```diff
@@ -0,0 +1,115 @@
...
+    UpdateUserMutationVariables
+  >(gql(updateUser));
+  useModalShortcuts({ onClose });
+
+  const createChangeHandler = (field: EditableField) => (
+    e: React.ChangeEvent<HTMLInputElement>
+  ) => {
+    setPristine(false);
+    setUserEdit({ ...userEdit, [field]: e.target.value });
+  };
```
alex-sf, line 50: How would you go about memoizing this curried function if we wanted to do a performance optimization that stoped this from breaking the shallow equality check in the components?

src/components/Mapbox.ts
```diff
@@ -0,0 +1,91 @@
...
+          mapboxRef.current?.setCenter(location.center);
+          markerRef.current?.setLngLat(location.center);
+          if (location.bbox) {
+            mapboxRef.current?.fitBounds(location.bbox);
+          } else {
+            mapboxRef.current?.setZoom(13);
+          }
+        }
+      } catch (e) {
+        captureException(e, { extra: { query } });
```
alex-sf, line 81: great sentry handling 🎉

pages/_app.tsx
```diff
@@ -0,0 +1,26 @@
+import { ApolloProvider, NormalizedCacheObject } from "@apollo/client";
+import { ThemeProvider } from "@emotion/react";
```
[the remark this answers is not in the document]
gztomas: Sure! I think both `styled-components` and `emotion` would have done a perfect job here, so there isn't really a need for `emotion`. I actually have been using `styled-components` for previous projects. And migrating between these is quite simple due to a very similar API.

However, I also have been using `material-ui` in the past and was aware of their migration to some other styling solution. At first, they were considering `styled-components`, but in the end, they went for `emotion`.

This issue describes the thorough analysis they did for making that choice https://github.com/mui-org/material-ui/issues/22342

Maybe the most interesting buy points from there in favor of emotion are concurrent mode and SSR out of the box.

On top of that, I do like `material-ui` and have experience using it, if I wanted to add MUI to this project, using the same styling solution they will use will give a smaller bundle size as opposed to adding something different, like `styled-components`

src/components/UsersList.tsx
gztomas: These articles from [Ryan Florence](https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578) and [KCD](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse) describe very well the reasoning behind using an inline function here.

Shallow props equality checks only happen if `UserCard` was a memoized component (or was using `shouldComponentUpdate`, as connected redux components do), otherwise, when the parent renders, the child renders, no matter if the props values change or not.

So, for this case, I believe, making the click handler stable across renders will not only fail to slightly improve performance, but quite the opposite will slightly degrade performance. Like KCD describes in his article, it would just add the extra cost of memoizing the functions, while the number of times `UserCard` renders will be the same exactly. The worst part is that to really know if memoizing the function is actually improving anything, this component will need to know if `UserCard` is really memoized or not, which is actually an implementation detail of another component.

I agree however if `UserCard` was an expensive component and if it was using `React.memo`, changing `UsersList` by making this click handler stable would work as an escape hatch for fixing a possible performance issue in`UserCard`.

I would try first though, to avoid changing this component due to performance issues in another component. That practice might end with a codebase full of `useCallback` and `useMemo` everywhere. I would instead favor other strategies, like making the render of `UserCard` more performant (rather than focusing on how many times it renders). That would depend on the expensive component implementation, but one way of achieving that could be by using composition in our favor to extract the expensive part out from `UserCard` (assuming that part does not need the click handler).

`<UserCard theExpensiveThing={theExpensiveThing} onClick={() => setSelected(item)}/>`

And then moving the expensive thing as up in the tree as possible, so it will be re-rendered as few times as possible, and `UserCard` won't be expensive anymore because it won't render the expensive thing.

This is explained much better in this other [article](https://kentcdodds.com/blog/optimize-react-re-renders/) from KCD.

Does all this sound right to you?

src/navigationState.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (limit) {
+    params.set("limit", String(limit));
+  } else {
+    params.delete("limit");
+  }
+  if (query) {
+    params.set("query", query);
+  } else {
+    params.delete("query");
+  }
```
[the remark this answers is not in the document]
gztomas: For this specific piece of code we could do:

```ts
const navigationState = { limit, query };
for (const [key, value] of Object.entries(navigationState)) {
  if (value) {
    params.set(key, value);
  }
  else {
    params.delete(key);
  }
}
```

I tend to follow the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) before applying a generalization. This is a very simple case, but in general, being explicit could be beneficial until the right generalization becomes evident.

src/components/UserEditor.tsx
gztomas: Please check my previous comment, my understanding is that making these handlers stable won't affect performance positively in this case.
If we however wanted to make these handlers stable we could use `lodash/memoize`:

```ts
  const createChangeHandler = React.useMemo(
    () =>
      memoize(
        (field: EditableField) => (e: React.ChangeEvent<HTMLInputElement>) => {
          setPristine(false);
          setUserEdit((userEdit) => ({ ...userEdit, [field]: e.target.value }));
        }
      ),
    []
  );
```

`useMemo` alone won't work here, since that would only make the `createChangeHandler` function stable, but what we actually need here is the return values of that function to be stable. That's what `memoize` provides. `useMemo` is needed as well though, otherwise we will be creating a different memoized function on every render.

Something important to note there is that it uses now the function form of `setUserEdit`, otherwise we would get stale references to `userEdit` if we were getting it from the closure like we were doing in the non-memoized alternative.

Besides all this, I would try to avoid this new implementation because of being too complex for this simple use case. The creator function is probably not worth the extra complexity in this case.

src/mockServer.ts
```diff
@@ -0,0 +1,86 @@
...
+  rest.get("https://source.unsplash.com/*", async (_, res, ctx) => {
+    const imageBuffer = await (
+      await fetch(PLACEHOLDER_AVATAR_IMAGE)
+    ).arrayBuffer();
+    return res(
+      ctx.set("Content-Length", imageBuffer.byteLength.toString()),
+      ctx.set("Content-Type", "image/jpeg"),
+      ctx.body(imageBuffer)
+    );
+  }),
```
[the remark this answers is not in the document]
gztomas: I'm not quite sure this answers the question, but I believe in that case these mocks can be a project by themselves, with some dir structure mapping the routing. If we had many services, they might even live next to the service (maybe same repo) they mock. So the team that owns that service can also provide mocks for it.
Or maybe the concern is more around the tool we are using here? Happy to provide more details.

src/components/UsersList.tsx
alex-sf: Awesome, this was exactly what I was looking for. Like I was saying, it doesn't really make a difference in this application since there will be no meaning performance boost by changing anything, but I wanted to have a discussion on the general principle of passing in anonymous functions into more expensive components. I think you hit the nail on the head with implementation strategies if we were in the scenario that I had brought up.

PR #26, opened 2020-12-01
[Test] Cassapula

cassapula:
## Overview
This is my solution to the Fullstack challenge. Thank you again for this opportunity. I'm looking forward to your feedback.

## Tech stack
This project was developed using Angular and NestJS with AWS (Lambda, API Gateway, S3, and DynamoDB).

## More information
All necessary information (such as demo, documentation, design, etc) are available on the `README.md` files inside both `api` and `web-app`.


web-app/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+This is part of the effort to keep the application scalable and extensible. Large applications could request too much data without the need, increasing the infrastructure cost.
+
+# Error handling & Logging 
+The application contains an interceptor (`core/error/interceptors/http-error.interceptor.ts`) responsible to catch all HTTP (backend) related errors and retry. If the interceptor didn't succeed in the second (or X) attempt, the error is automatically caught by our Error Handler (`core/error/error.ts/`) that logs the error to Sentry.
+
+# Design and user experience
+The application has its layout constructed with `SCSS` on the 7-1 SASS concept. The structural change only change was that the files related to specific pages are inside angular components. The scss files are imported into the main style.ts that is configured to be rendered inside the `angular.json`.
+
+Besides creating global styles (for buttons, avatar, card), the application also contains a grid system that works with breakpoints based on the viewport.
```
marcin-piela-sf, line 109: Have you tested how this "gird" behaves on a mobile device?

marcin-piela-sf (commented):
@cassapula Thanks for the PR! :)

Posted a few comments/questions, feel free to ask questions if anything is unclear.

web-app/src/app/core/users/services/users-query.service.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  private setUsers(users: User[]): void {
+    this.$users.next(users);
+  }
+
+  public getUsers(): Observable<Array<User>> {
+    return this.$users.asObservable();
+  }
+
+  public createUser(): void {
```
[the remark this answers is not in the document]
cassapula: Since I didn't find anything related to the user creation on the UI, I created this function that inserts a user on the DynamoDB when you reload the page using a special variable on the URL query params (`createUser=true`).

I forgot to add this to the documentation. The reason behind this was to make it easier to test the application in a different AWS environment.

web-app/src/environments/environment.prod.ts
```diff
@@ -0,0 +1,5 @@
+export const environment = {
```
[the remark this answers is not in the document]
cassapula: For Angular I believe it is, it's the recommended practice by the Angular team because it integrates with Angular CLI that automatically replaces the files with the appropriate version for the intended target environment.

So when deploying the application to production, the `environment.ts` content will be replaced with the content of the `enviromnent.prod.ts`.

web-app/src/app/core/layout/config/layout.config.ts
```diff
@@ -0,0 +1,30 @@
+import { Layout } from 'src/app/layout/layout.types';
+
+// Theme type
+export type Theme = 'light' | 'dark';
```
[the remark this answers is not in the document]
cassapula: It's not something that is used on the app at the moment, but the reason behind this was part of the effort of making the code extensible and scalable. Different types of layouts can be fundamental in the long-term of a large web application when we need to add a different layout for a specific page, or maybe make the code accessible (eg. color-blind mode).

web-app/e2e/src/app.e2e-spec.ts
```diff
@@ -0,0 +1,23 @@
+import { AppPage } from './app.po';
+import { browser, logging } from 'protractor';
+
+describe('workspace-project App', () => {
```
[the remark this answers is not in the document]
cassapula: To be honest, I wasn't able to create the tests on time, so currently, on the application, it's not properly working as it should.

The test should instantiate the main route and check inside the root route if `.content span` equals a specific string. For this test specifically, I didn't change the expected response after developing the main route of the application.

The idea initially was to create tests that would ensure that the components are working correctly together (page users with users edit and users search) to validate the entire user feature cohesively.

web-app/src/assets/utils/loader.svg
```diff
@@ -0,0 +1,36 @@
+<svg width="100" height="100" viewBox="0 0 44 44" xmlns="http://www.w3.org/2000/svg" stroke="#000">
```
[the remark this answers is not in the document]
cassapula: Sorry! That's not what I intended. 

Usually what I do is to use a library called [svg-icon](https://github.com/ngneat/svg-icon) in combination with [svg-to-ts](https://github.com/kreuzerk/svg-to-ts) that directly displays SVG icons in the DOM.

The svg icons are registered on the core module, and it's available to be used on the application without the need of absolute paths, with several options out of the box, like sizing, viewports, etc.

Normally I kept the svg files in `assets/svg/`. For this specific case, I added just to validate the loading functionality and I forgot to change.

web-app/src/styles/abstracts/_functions.scss
```diff
@@ -0,0 +1,30 @@
+// -----------------------------------------------------------------------------
+// This file contains all application-wide Sass functions.
```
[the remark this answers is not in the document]
cassapula: I am not a huge fan of creating comments everywhere because I believe the code should be descriptive enough with good variable names and logic that anyone can read without the need for comments. 

For this case, using a specific pattern with scss, I normally like to let comments describing what to do in each file to make it easier for other developers to keep the code clean and organized.

web-app/src/styles/components/_avatar.scss
```diff
@@ -0,0 +1,9 @@
+// -----------------------------------------------------------------------------
+// This file contains all styles related to the button component.
+// -----------------------------------------------------------------------------
+
+.avatar {
+  width: 168px;
```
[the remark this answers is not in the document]
cassapula: I am a big fan of utility classes. To handle this code better in the future I would remove the hardcoded width and leave it to each place to choose the specific width of the avatar using utility classes, or pre-define the width of the avatar by defining the width of the container that the avatar is inserted into.

web-app/src/app/modules/pages/users/users.component.ts
```diff
@@ -0,0 +1,118 @@
...
+    // Smooth scroll to bottom
+    setTimeout(() => {
+      window.scrollTo({ top: document.body.scrollHeight, behavior: 'smooth' });
+    });
+  }
+
+  private async getUsers(): Promise<void> {
+    this.usersQueryService.getUsers()
+      .subscribe(users => {
+        if (users == null) { return; }
```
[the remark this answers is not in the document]
cassapula: Yes, but to make sure it's going to be type-safe it would be required to create a validation in the service when it puts the users inside the `Subject`. The `Subject` is the type of `Behavior` that is initialized with a null variable, I used `==` to make sure I would handle any null and undefined cases.

api/src/app.module.ts
```diff
@@ -0,0 +1,36 @@
...
+import { UserModule } from './user/user.module';
+import { LocationModule } from './location/location.module';
+
+export const dynamooseConfiguration = {
+  local: process.env.IS_DDB_LOCAL === 'true',
+  aws: { region: process.env.REGION }
+}
+
+export const graphQLConfiguration = {
+  autoSchemaFile: true,
```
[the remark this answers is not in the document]
cassapula: Even though schema-first usually reduces development time and creates a better-designed API, I find it easier to manage using code-first when we expect the schema to grow in complexity. I tried to create the application always thinking it could become large and complex over time.

web-app/src/styles/components/_card.scss
```diff
@@ -0,0 +1,18 @@
+// -----------------------------------------------------------------------------
+// This file contains all styles related to the button component.
+// -----------------------------------------------------------------------------
+
+.card {
+  width: 400px;
```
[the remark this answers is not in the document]
cassapula: Since I am using px, that is an absolute unit, seeing on mobile would make the object fixed and not change based on any other elements.

A better approach would be to use REM that is a relative unit. Or not set any value hardcoded at all, and make it proportional to any parent size. In my case, since I have a grid created, it could have the same result without the need to hardcode anything.

The reason for me doing it was because I copied the style from Figma and forgot to refactor.

web-app/src/app/modules/components/users/search-user/search-user.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(SearchUserComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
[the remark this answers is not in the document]
cassapula: It checks if a component was successfully created. 

the `toBeTruthy` source code is as following:

```
getJasmineRequireObj().toBeTruthy = function() {
  function toBeTruthy() {
    return {
      compare: function(actual) {
        return {
          pass: !!actual
        };
      }
    };
  }

  return toBeTruthy;
};
```

For this particular case, it would be the same as validating if the component is defined.

web-app/README.md
cassapula: Yes. Although the breakpoint works, the app is not responsible because of the other elements, such as the card component.

web-app/src/styles/abstracts/_variables.scss
```diff
@@ -0,0 +1,92 @@
...
+// Light background color
+$light-background-color: rgb(248, 248, 248, 1) !default;
+
+/// Main brand color
+/// @type Color
+$brand-color: rgb(229, 0, 80) !default;
+
+/// Light grey
+/// @type Color
+$light-grey: rgb(237, 237, 237) !default;
```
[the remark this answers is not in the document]
cassapula: I created a few variables to personalize the application and define a theme at first, but since the application is simple I didn't continue to use it. However, it could be used on the few components we have such as the button component. It was a lack of attention on my part.

web-app/src/styles/components/_button.scss
```diff
@@ -0,0 +1,34 @@
+// -----------------------------------------------------------------------------
+// This file contains all styles related to the button component.
+// -----------------------------------------------------------------------------
+
+.btn {
+  width: 280px;
```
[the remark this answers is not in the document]
cassapula: Not at all, not only that but also do not recommend the use of px.

A hardcoded width like this will directly impact the responsivity of the layout. This is a technical debt on the application. I initially copied the style from Figma and later forgot to change/refactor.

web-app/src/styles/components/_button.scss
```diff
@@ -0,0 +1,34 @@
...
+  height: 90px;
+
+  background: #FFFFFF;
+  border: 4px solid rgba(0, 0, 0, 0.1);
```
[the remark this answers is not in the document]
cassapula: Definitely. We could simply have created the variables 

```
$light-color: #FFFFFF
$border-dark: rgba(0, 0, 0, 0.1)
```

And then used this instead

```
background: $light-color;
border: 4px solid $border-dark;
```

web-app/src/styles/vendors/_normalize.scss
```diff
@@ -0,0 +1,346 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
```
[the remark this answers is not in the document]
cassapula: It definitely could have. The only reason it was not is that I was following the 7-1 sass architecture and decided to leave it as a local file to exemplify the use of the pattern for the code review. But definitely for a real-world application, it would be better to have it installed and versioned with npm.

web-app/src/app/modules/components/dialog/dialog.component.scss
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: auto;
+  margin-left: auto;
+  min-height: 200px;
+  width: 100%;
+  max-width: 75%;
+  background-color: #fff;
+  box-shadow: 0 7px 8px -4px rgba(0, 0, 0, 0.2), 0 13px 19px 2px rgba(0, 0, 0, 0.14), 0 5px 24px 4px rgba(0, 0, 0, 0.12);
+}
+
+@media (min-width: 768px) {
```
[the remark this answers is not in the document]
cassapula: It would be easier and more inclusive since we want to cover all of our users.

I could have created a mixin that accepts parameters and selects the desired breakpoint

```
@mixin for-size($size) {
  @if $size == phone-only {
    @media (max-width: 599px) { @content; }
  } @else if $size == tablet-portrait-up {
    @media (min-width: 600px) { @content; }
  } @else if $size == tablet-landscape-up {
    @media (min-width: 900px) { @content; }
  } @else if $size == desktop-up {
    @media (min-width: 1200px) { @content; }
  } @else if $size == big-desktop-up {
    @media (min-width: 1800px) { @content; }
  }
}
```

And then to use it like this:

```
.dialog {  
   top: 6rem;  
 
   @include for-size(phone-only) {    
      top: 3rem; 
   }
}
```

web-app/src/app/core/users/services/users-query.service.ts
```diff
@@ -0,0 +1,146 @@
...
+
+        this.apollo.mutate({ mutation: createUsers }).subscribe();
+      });
+  }
+
+  public listUsers({ page = 1, refetch = false }): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+      const queryStmt = gql`
+      {
+        listUsers(
```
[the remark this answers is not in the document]
cassapula: Usually, the queries could be created and exported in a `graphql.ts` file inside `src/app`, and then I could merely import the queries and mutations into the components or services when needed.

web-app/src/app/data/model/user.ts
```diff
@@ -0,0 +1,9 @@
+export interface User {
```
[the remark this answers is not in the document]
cassapula: I could potentially use the graphql code generator to generate a strongly type angular service for the defined queries, subscriptions, and mutations. For that we need to simply create a `.graphql` file to write our queries and use `graphql-codegen` to generate the angular services, it has benefits on doing that, for instance, allowing FE and BE developers to work at the same time and test the client independently of having the server ready, and save development time.

web-app/src/app/modules/components/users/edit-user/edit-user.component.ts
```diff
@@ -0,0 +1,58 @@
...
+      .pipe(debounceTime(1000))
+      .subscribe(async address => {
+        if (address) {
+          this.usersEditService.geoLocation = await this.locationService.getLocationByAddress(address);
+        }
+      });
+  }
+
+  public async updateUser(): Promise<void> {
+    this.usersService.isLoadingData = true;
```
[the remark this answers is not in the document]
cassapula: That is a complicated question, but my answer for 99.9% of the cases is no.

Angular is a complete framework and offers alternatives, equal, or even superior built-in tools to deal with the problems that lead us to use such libraries on React, for example.

Per the official Angular documentation, one way of communication between components is using a shared service. The Hierarchical Dependency Injection makes it possible to define the scope of service visibility and make it visible only to the consuming hierarchy of the components.

Also, features from Angular Router, such as child and auxiliary routes, and others support URL reactively, and we can use query parameters as observables, making all subscribers be notified on a chance.

Also, Angular strictly enforces unidirectional data flow on the presentation layer (via input bindings), and prevents the view from updating itself out of the box.

So for this case, my answer will be no, unless we're dealing with hundreds of thousands of users and want to spare a little server resource by syncing state on the client, I believe it's not necessary and maybe even dangerous. But if I need to use, I would recommend NgXs that feels natural to the angular ecosystem (by incorporating TS decorators)

web-app/src/app/modules/pages/users/users.component.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        if (this.usersService.queryParamPage.page > 1) {
+          this.loadMoreDisabled = users.length < (this.usersService.queryParamPage.page * PageInput.LIMIT);
+        }
+
+        this.usersService.users = users;
+        this.usersService.isLoadingData = false;
```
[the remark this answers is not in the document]
cassapula: That would be acceptable per the [official Angular docs](https://angular.io/guide/component-interaction#parent-and-children-communicate-via-a-service) that state that one way of communication between components is the use of a shared service. I have created the `UsersService` specifically to manage the communication between users' components.

Another way to do this would be to create a two-way data binding using inputs and outputs, but that would create unnecessary logic.

web-app/README.md
```diff
@@ -0,0 +1,149 @@
...
+- Caching
+- Application configuration
+
+The required features must be declared and exported on the main core module `core.module.ts` to have the dependency injected on the main angular module.
+
+## Data
+Structural folder to hold typed objects, interfaces, enums, and others.
+
+## Caching
+Caching is a functionality added on `core/cache` that is responsible to identify similar requests, and if required, retrieve the response in-memory instead of calling the API, unless it's excluded on `core/cache/interceptors/http-cache.interceptor.ts`.
```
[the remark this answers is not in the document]
cassapula: You're right. Using Apollo Cache would be a better approach, unless if we're dealing with also a lot of calls to RESTful APIs, but that's not the case. This should be removed to free space for apollo in-memory cache.

cassapula:
Hey @marcin-piela-sf! Thank you so much for reviewing my PR! I responded to your comments, but please, feel free to ask any questions if something wasn't clear.

api/package.json
```diff
@@ -0,0 +1,124 @@
...
+    "genres": "ts-node genres.ts -- \"src/**/*.schema.ts\" \"resource/dynamodb.yml\""
+  },
+  "dependencies": {
+    "@graphql-tools/merge": "6.2.5",
+    "@mapbox/mapbox-sdk": "^0.11.0",
+    "@nestjs/common": "7.5.5",
+    "@nestjs/config": "0.6.1",
+    "@nestjs/core": "7.5.5",
+    "@nestjs/graphql": "7.8.2",
+    "@nestjs/platform-express": "7.5.5",
```
marcelino-sf, line 37: What are your thoughts on using a framework for building a lambda to only manage contacts?

api/package.json
```diff
@@ -0,0 +1,124 @@
...
+    "@types/mapbox__mapbox-sdk": "^0.6.5",
```
marcelino-sf, line 38: Can you help me understand why this `@types` is necessary to be here instead of `devDependencies`?

api/serverless.yml
```diff
@@ -0,0 +1,59 @@
+service: superformula-api
```
marcelino-sf, line 1: I checked the generated bundle and it has 42MB, when you're creating a new lambda, how do you balance that?

api/serverless.yml
```diff
@@ -0,0 +1,59 @@
...
+  name: aws
+  runtime: nodejs12.x
+  stage: ${opt:stage, 'dev'}
+  region: ${opt:region, 'us-east-1'}
+  environment:
+    SERVICE: ${self:service}
+    STAGE: ${self:provider.stage}
+    REGION: ${self:provider.region}
+    NO_COLOR: ${env:NO_COLOR, 'true'}
+    MAPBOX_API: ${env:MAPBOX_API, 'sk.eyJ1IjoiYnJ1bm9jYXNzIiwiYSI6ImNraTU2cHBrNjA4YjMyc21zdmJqeDRkYXYifQ.ZMnq8HGnNmRRIpC4Zz6UlA'}
```
marcelino-sf, line 13: How do you manage environment variables in a real-world project? And what are your thoughts on pushing secrets inside code?

api/src/user/controller/user.controller.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const spy = jest.spyOn(UserService.prototype, 'create')
+      .mockImplementation(() => {
+        return userInput as any
+      })
+
+    const result = await controller.create(userInput);
+    expect(result).toBe(userInput)
+
+    spy.mockRestore()
```
marcelino-sf, line 41: Can you help me understand the reason you're calling this here instead of an `afterEach`?

api/src/user/controller/user.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+  Post
+} from '@nestjs/common';
+
+import { UserService } from '../service/user.service';
+import { CreateUserInput } from '../model/create-user.input';
+import { UpdateUserInput } from '../model/update-user.input';
+import { GetUserInput } from '../model/get-user.input';
+import { DefaultQueryInput } from '../../common/model/query.input';
+
+@Controller('user')
```
marcelino-sf, line 18: Can you help me understand why you decided to create a REST and also a GraphQL API? When would you opt for REST over GraphQL and otherwise?

api/src/user/model/update-user.input.ts
```diff
@@ -0,0 +1,25 @@
...
+  @IsString()
+  @Field()
+  description: string;
+
+  @IsNotEmpty()
+  @IsString()
+  @Field()
+  address: string;
+
+  // TODO: if we want to implement a logic/soft delete we can toggle the status between active and inactive
```
marcelino-sf, line 22: How do you manage TODOS and technical debts in real-world projects?

api/src/user/service/user.service.ts
```diff
@@ -0,0 +1,78 @@
...
+import { DefaultQueryInput } from 'src/common/model/query.input';
+
+@Injectable()
+export class UserService {
+  constructor(
+    @InjectModel('user')
+    private readonly model: Model<User, UserKey>,
+  ) {}
+
+  public async create(input: CreateUserInput) {
```
marcelino-sf, line 20: What is the difference between declaring a function as `public` instead of just declare the function without the `public` term?

api/src/user/service/user.service.ts
```diff
@@ -0,0 +1,78 @@
...
+  ): Promise<ScanResponse<User> | ErrorResponse> {
+    try {
+      if (input.page) input.limit = input.page * input.limit;
+
+      const response = await this.model.scan().limit(input.limit).exec();
+
+      return response;
+    } catch (error) {
+      console.log(error);
+      return { error: true, msg: error.message };
```
marcelino-sf, line 52: Can you help me understand the reason you chose to return an ErrorResponse instead of throwing an Error?

marcelino-sf (commented):
Hey @cassapula, thank you for your time writing this PR. I left a few comments and I'll be happy to see your thoughts. Cheers!

api/src/location/service/location.service.ts
```diff
@@ -0,0 +1,35 @@
...
+export class LocationService {
+  readonly geocodeService: GeocodeService;
+
+  constructor() {
+    this.geocodeService = mapboxGeocodeServiceFactory({
+      accessToken: process.env.MAPBOX_API as string,
+    });
+  }
+
+  public async getGeocodeCoordinates(address: string) {
```
marcelino-sf, line 16: What are your thoughts of adding a return type to this and other functions?

api/src/location/service/location.service.ts
```diff
@@ -0,0 +1,35 @@
...
+      })
+      .send();
+
+    const feat = location.body?.features;
+    const coordinates = feat && Array.isArray(feat) && feat[0]?.center;
+    const coordinatesFound = coordinates?.length === 2;
+    if (!coordinatesFound)
+      throw (
+        'Did not found any coordinate for the following address: ' + address
+      );
```
marcelino-sf, line 31: What are your thoughts on throwing String as opposed to an Error class?

api/src/location/service/location.service.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable } from '@nestjs/common';
+import mapboxGeocodeServiceFactory, {
+  GeocodeService,
+} from '@mapbox/mapbox-sdk/services/geocoding';
+
+@Injectable()
```
marcelino-sf, line 6: What are your thoughts on using `inversion-of-control`/`dependency-injection` design-pattern inside a lambda function?

api/src/user/service/user.service.ts
cassapula: The default access level for a member is public unless we specify otherwise. I believe it's a matter of personal preference. 

If you use parameter properties though, the explicit public access modifier is compulsory in order to create and initialize an instance member from the given parameter.

I prefer to use it for two main reasons
-  It shows that the class was intended to be public
- Makes it more clear for people that come from other languages such as C# (that the default modifier is private).

api/src/user/model/update-user.input.ts
cassapula: I believe that only writing TODOs along the code can be chaotic while the project increases. If we have several repositories, the technical debt will start to be impossible to collect.

For a real-world application, I believe we need a bit more information to disclaimer our technical debt. Such as cause, possible solution, and a ticket number so we can track the technical debt together with other issues and bugs and our backlogs.

It can take more time, but it's a small price to pay for control of the technical risks, and understand when we need a sprint for code stabilization since technical debt is something common and normal to happen.

api/src/user/controller/user.controller.ts
cassapula: I included it because it's easy to implement from the NestJS perspective, but it's not being used anywhere. There's no reason to have it both for the same endpoint, but to make the code extensible I choose to implement, to serve as an example in a scenario where we want to use rest.

I believe GraphQL is superior in most scenarios because it solves problems that REST suffered for years, with GraphQL we can retrieve all the initial data in a single roundtrip to the server and reduce the dependency of the client on the server.

A scenario that I would choose REST is when creating an API for analytics, I believe it's easier to obtain with REST, due to the limited amount of tooling for GraphQL.

api/src/user/controller/user.controller.spec.ts
cassapula: Oops. You're right! To be honest, there's no reason, I should have called it in the afterEach.

api/src/user/service/user.service.ts
cassapula: Sorry, my mistake! - I added this `try/catch` block while I was developing and I faced an unrelated issue, to help me debug, and I forgot to remove it.

api/src/location/service/location.service.ts
cassapula: For my specific scenario, may be okay because it's very simple, but in a real-world application it's a bad practice because the catch can be outside of our code, and in this case, it might expect a full Error instance and not a plain string.

api/serverless.yml
cassapula: Secrets and tokens should be safe storage, and also should be easier to be retrieved by our app.

For my case, I am passing this sensitive data as environment variables, but in a real-world app, I believe we need to do more.

One solution that I frequently am using is Vaults (more specific, HashiCorp Vault), which is a light solution with a comprehensive API that sotrage the secrets in a safe and certificated manner and uses strict access controls to recover the data when needed.

api/src/location/service/location.service.ts
cassapula: Usually, TypeScript can figure the return type out by looking at the return statements, so we can also optionally leave this off in many cases. The TypeScript compiler will infer types when it can.

But in other cases, it will cause an error because the compiler with the given statements has no way to decide the return type, and this is where we have to exploit the return type, for example:

```
foo() {
    if (true)
        return 'bar'
    else 
        return 0
}
```

So in some places is required and others are not. I am all about consistency so I believe we should use it everywhere if that's the guidelines of the code we're following.

api/src/location/service/location.service.ts
cassapula: Well, since the dependency injection will provide objects and dependencies that our lambda needs, delegating the instantiation of dependencies to the IoC container (in my case, the NestJS runtime system), instead of doing it in the code imperatively.

I believe we can have a lot of benefits doing so, like parallel development on the dependencies, easily swap in mock objects for testing purposes, enabling loosely-coupled design and etc.

api/package.json
cassapula: I enjoy the dependency injection that comes with Nest.js because it let me focus on building and abstracting out providers and services and focus on my application's logic. Also, If you have a preference for TypeScript over JS and want to build a GraphQL server, Nest + TypeORM is a good option, with the code first approach, we can use decorators to generate the GraphQL schema automatically from code, and mapping input types to TypeScript classes makes for easily readable and maintainable code.

But the point is that even though the application is simple now, I followed principles to make it scalable, and once the application grows, the use of the framework will pay off.

api/package.json
cassapula: I took a look in my code again and I believe it doesn't have to be here and could be moved to the `devDependencies`.

The only reason for keeping it in here is if we're exporting the type from `@types/mapbox__mapbox-sdk`, then Typescript consumers of my package will be unable to guess what the exported type from `@types/mapbox__mapbox-sdk` is since devDependencies of my package are not installed.

But, it's not the case here!

api/serverless.yml
cassapula: That's a good question because for this app I didn't check the bundle size before creating this PR. 

But I should have reviewed the project to make it as clean as possible, and to only bundle what's needed for the application to run, making sure to remove all devDependencies, docs, tests, scripts, and other files that might be in the repository and it's not necessary for the production app - after deployment and CI/CD routines.

cassapula:
Hey @marcelino-sf! Thank you so much for reviewing my PR! It was awesome this collaboration to see all potential issues, risks, and areas of improvement on this app. 

I already responded to your comments, and I am sorry for the delay in doing that. Feel free to ask any questions if something wasn't clear. Cheers!

PR #28, opened 2020-12-11
Ezequiel Garcia Solution

ezequielgarcia88:
Hello Guys,

Here is my PR request for the code test.
You can try the FE online demo [here](http://test-full-stack-frontend.s3-website.us-east-2.amazonaws.com/) and the BE lives in [here](https://5sjfvum52nhujbxdurjdnjlw6m.appsync-api.us-east-2.amazonaws.com/graphql).

Let me know if you have any questions or concerns.
Thanks again for the opportunity!!


frontend/src/components/UserModal/UserModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+    name: '',
+  });
+  const handleUserPropertyUpdate = (property: string, newValue: string) => {
+    setUserInModal({
+      ...userInModal,
+      [property]: newValue,
+    });
+  };
+
+  useEffect(() => setUserInModal(user || userInModal), [JSON.stringify(user)]);
```
mat-sf, line 27: Can you elaborate more on why did you use `JSON.stringify(user)` here?

frontend/src/components/App/App.tsx
```diff
@@ -0,0 +1,119 @@
+import React, { useCallback, useState } from 'react';
+import { useQuery, useMutation } from '@apollo/client';
+import { debounce } from 'lodash';
```
mat-sf, line 3: What do you think about importing lodash this way? How about `lodash.debounce`?

frontend/src/components/Button/Button.module.scss
```diff
@@ -0,0 +1,34 @@
+.button {
+  background-color: transparent;
+  border: 4px solid rgba(0, 0, 0, 0.1);
+  border-radius: 8px;
+  box-sizing: border-box;
+  color: #000;
+  cursor: pointer;
+  font-family: 'Source Sans Pro', sans-serif;
+  font-size: 24px;
```
mat-sf, line 9: Do you know any other units for font-sizes? Why you decided to use px?

frontend/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import classnames from 'classnames';
+import styles from './Button.module.scss';
+
+type Props = {
```
mat-sf, line 5: Why type instead of interface (I'm curious, I know that there is no right answer to this 😄)?

frontend/src/components/Card/Card.module.scss
```diff
@@ -0,0 +1,62 @@
+.wrapper {
+  background-color: #fff;
```
mat-sf, line 2: How would you share such values across the project (`#fff`, border radiuses etc.)?

frontend/src/components/App/App.tsx
```diff
@@ -0,0 +1,119 @@
...
+      refetch();
+    }, SEARCH_DEBOUNCE_TIME),
+    [],
+  );
+  const handleUserSelection = (user?: User) => {
+    setSelectedUser(user);
+    document.body.style.overflow = user ? 'hidden' : '';
+  };
+
+  console.log('zeta', loading, error, data);
```
mat-sf, line 52: What do you think about having `console.log`s in prod app?

frontend/src/components/App/App.tsx
```diff
@@ -0,0 +1,119 @@
...
+              setSearchsText(event.target.value);
+              setSearchHandler(event.target.value);
+            }}
+          />
+        </div>
+      </header>
+      <div className={styles.list}>
+        {users.map((user) => (
+          <Card
+            key={user.id}
```
mat-sf, line 78: Let's assume you don't have unique value in user object, would you use array index as key?

frontend/cypress/integration/spec.js
```diff
@@ -0,0 +1,6 @@
+describe('Test Full Stack', () => {
+  it('shows header and searh input', function () {
```
mat-sf, line 2: What scenarios would you test in e2e? I'm asking because I cannot see any more e2e tests

mat-sf (commented):
Hi Ezequiel! I left a few comments where we can have a little discussion, if you have any questions, feel free to ask.

Few more questions from me (frontend only):
1. How would you make that app more accessible?
2. How would you make that app mobile-friendly if we would provide you designs?
3. Would you improve something here if you would have more time?

BTW Good job! We appreciate the time you have spent on this.

chimon2000:
When I attempt to view an entry I see the following.

Screen Shot 2020-12-11 at 1 58 02 PM

Changing the address doesn't seem to resolve this.

frontend/src/components/App/App.tsx
ezequielgarcia88: Ups I missed this one when checking my code.

ezequielgarcia88:
@chimon2000 I was not able to add that functionality before sending the PR. I didn't want to take much longer to send it so you guys can start reviewing it.

I have added it, check the [online demo](http://test-full-stack-frontend.s3-website.us-east-2.amazonaws.com/).

frontend/src/components/UserModal/UserModal.module.scss
```diff
@@ -0,0 +1,132 @@
...
+.map {
+  background-color: blue;
+  border-radius: 8px;
+  flex-shrink: 0;
+  height: 336px;
+  margin-top: 5px;
+  width: 518px;
+}
+
+.rightPanel {
```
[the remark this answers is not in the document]
ezequielgarcia88: I normally follow the [BEM naming convention](http://getbem.com/naming/) for CSS classes. Probably messed it up in some places.

frontend/src/components/UserModal/UserModal.test.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+import { render, fireEvent, screen } from '@testing-library/react';
```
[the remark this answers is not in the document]
ezequielgarcia88: To be honest I just used it cause it was included when I created the APP with the CRA.
But the main difference is that `react-testing-library` is not focused on the component implementation details, you can not access the state, props and children components of a React component. Basically, when using `react-testing-library` the testing is done from the user’s perspective.

frontend/src/components/UserModal/UserModal.module.scss
```diff
@@ -0,0 +1,132 @@
+.modal {
+  display: table;
```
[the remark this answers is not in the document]
ezequielgarcia88: I used to basically center the modal content horizontally and vertically. Instead of it, I could have used the following for the backdrop:
```css
.backdrop {
  align-items: center;
  background-color: rgba(0, 0, 0, 0.3);
  display: flex;
  height: 100%;
  justify-content: center;
  width: 100%;
}
```

frontend/src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import moment from 'moment';
```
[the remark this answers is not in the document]
ezequielgarcia88: I choose `moment` because it is my default lib for handling dates, probably it is an overkill in this scenario since we only need to format some dates.
I could have used simpler and lighter like luxon, date-fns, day.js; or simply create some vanilla js code to do the format ([like](https://stackoverflow.com/questions/3552461/how-to-format-a-javascript-date)).

frontend/src/components/UserModal/UserModal.tsx
ezequielgarcia88: I normally use the `JSON.stringify` to check if any of the object properties has changed. In this case, I could have just made the comparison with the User ID.

frontend/src/components/Button/Button.module.scss
ezequielgarcia88: I simply used px because thats how the font-sizes are described in the [fimga](https://www.figma.com/file/hd7EgdTxJs2fpTzzSKlNxo/Superformula-full-stack-test). I'm familiar with other units like pt, em, vw, vh, vmin, vmax, etc.

frontend/src/components/App/App.tsx
ezequielgarcia88: I should have imported it like `import debounce from 'lodash/debounce'` to get a smaller bundle size:
Screenshot 2020-12-11 at 20 02 47

frontend/src/components/Card/Card.module.scss
ezequielgarcia88: Good point! I should have created a variables file and import them into the other scss files.

frontend/src/components/Button/Button.tsx
ezequielgarcia88: Yeah I am back and forth on this topic lately 😃 
I know interfaces are a little more powerful and flexible than type, but I think that most of the cases you would come across when coding with React can be expressed with type. Some cases where I would consider interfaces are when you do declaration merging of prop types in the global scope, you want to hide type implementation details, etc.

frontend/src/components/App/App.tsx
ezequielgarcia88: I would definitely not, because it doesn't uniquely identify the elements. In cases where the array is sorted or an element is added to the beginning of the array, the index will be changed even though the element representing that index may be the same. This results in unnecessary renders.

frontend/cypress/integration/spec.js
ezequielgarcia88: Yeah, I'm sorry about this, I run out of time to add more e2e scenarios and I didn't want to take much longer before submitting the PR. I would probably have added the following scenarios:
- should display the list of users.
- should be able to load more users to the list.
- should be able to filter the user's list.
- should be able to edit a user in the modal.
- should display the location in the map when the user address is changed.

ezequielgarcia88:
@mat-sf I answered all your inline questions. I will get back to you with the other 3 questions as soon as possible.

ezequielgarcia88:
@mat-sf Here are my answers to your other questions:

1. To make that App more accessible, I would have done the following actions:
    - Structure the App with Semantic Markup: use elements other than div, which can greatly improve accessibility and readability (like header, footer, section, aside, etc.).
    - Add ARIA Attributes: ARIA stands for Accessible Rich Internet Applications, and by adding them we make the App more readable for screen readers. We can also add the eslint plugin `eslint-plugin-jsx-a11y` to help with the accessibility rules on JSX elements.
    - Improve keyboard accessibility: we need to ensure that the App can be fully operated with the keyboard only. For example, we should let people navigate the App using just their keyboard (by using the `tabindex` attribute); or we should ensure that all functionality exposed through a mouse or pointer event can also be accessed using the keyboard alone.
    - Add Alt Text on Images: we can use the alt text to provide a description of the image’s content when screen reader would read out the App.
    - Improve focus management: since React applications continuously modify the DOM during runtime, sometimes leading to keyboard focus being lost or set to an unexpected element, we need to manually move focus in the right direction.
2. To make that App more accessible, I would have done the following actions:
    - 
    - 
    - 
How would you make that app mobile-friendly if we would provide you designs?
3. Of course, I would have. I would have improved on the following items:
    - Add all the e2e scenarios that I have described in my previous answer. And also have a mocked data set to run the scenarios.
    - Continue on adding more unit tests, in order to increase the code coverage.
    - Perform the actions I mentioned above to make the App more accessible.
    - Add Storybook integration with tests.
    - Properly handle the loading states when fetching the users, getting the user address location or modifying an user (by showing a loading indicator or something similar)
    - Inform when successfully editing an user by displaying an alert or notification.
    - Properly handle server errors by displaying an alert or notification.
    - Add the ability to delete users.
    - Add the ability to create new users.
    - Add sentry to monitor and track App errors or crashes.
    - Add form inputs validation.
    - Make the App mobile-friendly.
    - Inlcude GraphQL subscriptions support to get real-time updates of the User list from the server.

ezequielgarcia88:
@mat-sf Here are my answers to your other questions:

1. To make that App more accessible, I would do the following actions:
    - Structure the App with Semantic Markup: use elements other than div, which can greatly improve accessibility and readability (like header, footer, section, aside, etc.).
    - Add ARIA Attributes: ARIA stands for Accessible Rich Internet Applications, and by adding them we make the App more readable for screen readers. We can also add the eslint plugin `eslint-plugin-jsx-a11y` to help with the accessibility rules on JSX elements.
    - Improve keyboard accessibility: we need to ensure that the App can be fully operated with the keyboard only. For example, we should let people navigate the App using just their keyboard (by using the `tabindex` attribute); or we should ensure that all functionality exposed through a mouse or pointer event can also be accessed using the keyboard alone.
    - Add Alt Text on Images: we can use the alt text to provide a description of the image’s content when a screen reader would read out the App.
    - Improve focus management: since React applications continuously modify the DOM during runtime, sometimes leading to keyboard focus being lost or set to an unexpected element, we need to manually move focus in the right direction.
2. To make that App more mobile-friendly, I would perform the following actions:
    - Use the Browser Dev Tools to emulate different mobile devices on the browser, so we can properly develop the App to be displayed properly for a large number of mobile devices.
    - Use CSS3 Media queries to add breakpoints where certain parts of the App will behave differently on each side of the breakpoint. Since there are a lot of screens and devices with different heights and widths, it is hard to create an exact breakpoint for each device, so to keep things simple we could target a fixed number of groups.
    - Use the `viewport` meta tag to control the width, initial scale, maximum and minimum scalability, etc.
    - If we want to use Media Queries programmatically we can use libraries like `react-responsive`, `react-media-hook`, `react-responsive-picture`, etc.
3. Of course, I would have. I would have improved on the following items:
    - Add all the e2e scenarios that I have described in my previous answer. And also have a mocked data set to run the scenarios.
    - Continue on adding more unit tests, in order to increase the code coverage.
    - Perform the actions I mentioned above to make the App more accessible.
    - Add Storybook integration with tests.
    - Properly handle the loading states when fetching the users, getting the user address location, or modifying a user (by showing a loading indicator or something similar)
    - Inform when successfully editing a user by displaying an alert or notification.
    - Properly handle server errors by displaying an alert or notification.
    - Add the ability to delete users.
    - Add the ability to create new users.
    - Add sentry to monitor and track App errors or crashes.
    - Add form inputs validation.
    - Make the App mobile-friendly.
    - Include GraphQL subscriptions support to get real-time updates of the Users list from the server.

frontend/src/components/App/App.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const setSearchHandler = useCallback(
+    debounce((newSearch: string) => {
+      setSearch(newSearch);
+      refetch();
+    }, SEARCH_DEBOUNCE_TIME),
+    [],
+  );
+  const handleUserSelection = (user?: User) => {
+    setSelectedUser(user);
+    document.body.style.overflow = user ? 'hidden' : '';
```
alex-sf, line 49: It looks like this is to stop the scrolling for the modal if I'm not mistaken. Is there any other way you could have achieved this without doing a DOM manipulation like this?

frontend/src/components/App/App.tsx
```diff
@@ -0,0 +1,117 @@
...
+      handleUserSelection(undefined);
+    },
+  });
+  const setSearchHandler = useCallback(
+    debounce((newSearch: string) => {
+      setSearch(newSearch);
+      refetch();
+    }, SEARCH_DEBOUNCE_TIME),
+    [],
+  );
```
alex-sf, line 46: Is there any benefit to using a `useCallback` here?

frontend/src/components/UserModal/UserModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+    if (coordinates) {
+      const map = new mapboxgl.Map({
+        container: mapRef.current || '',
+        style: 'mapbox://styles/mapbox/streets-v11',
+        center: coordinates,
+        zoom: 10,
+      });
+      new mapboxgl.Marker().setLngLat(coordinates).addTo(map);
+    }
+  }, [JSON.stringify(location)]);
```
alex-sf, line 77: Was `useLazyQuery` not returning a consistent object to memoize this effect?

frontend/src/components/UserModal/UserModal.tsx
```diff
@@ -0,0 +1,149 @@
...
+      }
+    }, GET_LOCATION_DEBOUNCE_TIME),
+    [],
+  );
+
+  useEffect(() => {
+    const newUser = user || userInModal;
+    setUserInModal(newUser);
+    addressChangeHandler(newUser.address);
+  }, [JSON.stringify(user)]);
```
alex-sf, line 64: You shouldn't need to stringify the user for this check. That is telling me that the user object reference is changing too often. Do you know why that is?

ezequielgarcia88: I normally use the JSON.stringify to check if any of the object properties has changed. In this case, I could have just made the check with the User ID.

frontend/src/components/App/App.tsx
ezequielgarcia88: The idea was to memoize the debounced fn, so as not to recreate it in every render. I could have used useMemo.

frontend/src/components/UserModal/UserModal.tsx
ezequielgarcia88: I'm sorry, but memoize what effect?

frontend/src/components/App/App.tsx
ezequielgarcia88: Yeah, you are correct. To prevent DOM manipulation, I should have structured the app in such a way that the scrolling container is the App component instead of body (make the body overflow hidden) and have the App component disable or enable the scrolling through a conditional class.

PR #35, opened 2021-02-08
Full Stack Test

falcoagustin:
## Senior Backend Engineer - Agustin Falco

Greetings again! Thank you for letting me have the possibility to take this tests. It was long but totally worth it and also had a lot of fun, really appreciate it! 

I learnt a lot in the road, and thank you for taking the time to review what I've done!!

## Demo

This [link](http://fullstack-test-asset-bucket.s3-website-us-east-1.amazonaws.com/) contains a link for the Demo

## Notes

* There are things to be improved, as always (and thankfully 😄 ) as for example running some Linters, setting up CI/CD, sending the errors from the API into sentry by creating a middleware, etc. but I understand it as a task and the idea is to show how we work and what we know and that time is a constraint.
* I aimed to show how I worked and what I know, hence there are `TODOs` around the code that show something needs improvement for later, as when deciding for finishing the functionality vs doing something perfect, I chose finishing it.
* I tried to be as tidy as possible but still trying to achieve at least all basic requirements for both codebases, regardless that it was a BE-centered test.

## Documentation

Documentation is on `docs` folder for both of the projects, where I state everything asked that is asked and decisions taken along the process.

Thanks!!!


backend/__tests__/userMutations.test.js
```diff
@@ -0,0 +1,110 @@
...
+    }).toThrow(ValidationError)
+  });
+});
+
+describe('test delete user functionality', () => {
+  beforeEach(() => {
+    resetMockClient();
+  });
+
+  test('deletes user successfully', () => {
```
patrick-sf, line 103: what would a test for non-successful user deletion look like?

backend/__tests__/userQueries.test.js
```diff
@@ -0,0 +1,38 @@
...
+    }).toThrow(ValidationError);
+  });
+
+  test('gets users last key for pagination', () => {
+    const lastKey = getRandomHash();
+    setupMockClient('scan',{
+      Items: [ { id: getRandomHash() }, { id: lastKey }], LastEvaluatedKey: { id: lastKey }
+    })
+
+    return getUsers({ limit: 2 })
```
patrick-sf, line 33: could you please help me understand the reasoning for **limit: 2**?

backend/__tests__/utils.js
```diff
@@ -0,0 +1,27 @@
...
+
+const setupMockClient = (operation, data, err = null) => {
+  initializeClient(documentClient(operation, data, err))
+}
+
+const documentClient = (operation, data, rej) => {
+  return {
+    [operation]: () => ({
+      promise: () => {
+        if (!rej) {
```
patrick-sf, line 11: I noticed you used a ternary operation in the file below -- is there any reason why a ternary didn't seem appropriate here?

backend/common/dynamodb.js
```diff
@@ -0,0 +1,21 @@
+const AWS = require('aws-sdk');
+
+let client
+
+const initializeClient = (dynamoDB) => {
+  client = dynamoDB ? dynamoDB : new AWS.DynamoDB.DocumentClient()
```
patrick-sf, line 6: Just out of curiosity, what're your thoughts on ternaries? Are they beneficial? Easy to read? etc.

backend/serverless.yml
```diff
@@ -0,0 +1,46 @@
...
+provider:
+  name: aws
+  runtime: nodejs12.x
+  stage: dev
+  region: us-east-1
+  environment:
+    DYNAMODB_TABLE: 'Users'
+    MAPBOX_HOST: 'api.mapbox.com'
+    MAPBOX_PORT: 443
+    # Public access token for Mapbox API
```
patrick-sf, line 11: Would you handle this any differently if it were a private access token?

patrick-sf (commented):
How would you handle logging backend requests (both successful and failed)?

backend/__tests__/userMutations.test.js
falcoagustin: Greetings 😄 !
Great question! There is something to take on account on this specific tests, and it is that they are really coupled to the implementation of DynamoDB and how it works, therefore, the underlying behavior of it is either something we need to implement for unit testing. 

For the case when deleting a user by specifying a key of a user that doesn't exist, then it will return Success as well, providing no additional information _by default_ given that it is an idempotent operation, as per [here](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html).

We have two options the way I see it :) 
*Option 1:* If we now have a requirement for extra information for unsuccessful delete, the easiest thing to do change the change the current behavior and add a conditional expression on the query to check if it exists, and then catch the `ConditionalCheckFailedException` error, and create a mock behavior accordingly that throws that specific error:
```js
const params = {
  TableName: TABLE_NAME',
  Key: {
    id,
  },
  ConditionExpression: 'id = :val',
  ExpressionAttributeValues: {
    ':val': id
  }
};
```
*Option 2:* The other option that we have is to check if it that user exists after we executed the operation, given that delete is done on a key basis which is unique and idempotent operation, there is no risk on race conditions. Then we can mock that post-operation and retrieve the user with the same id, meaning that it's still there and wasn't deleted.

For other cases like connection times out, or other reasons, creating a mock that throws an error should be sufficient to see how it behaves by returning the mocked error.

backend/__tests__/userQueries.test.js
falcoagustin: Indeed 😄 
My reasoning behind `limit: 2` is that I wanted to test the pagination on a collection that does not contain a single user, even though the test would have been really similar, only changing which key is the one returned. Creating a test with multiple other users could have been a possibility as well. Now that I see it, this can be improved by using the last value of the collection, therefore:

```js
...
const lastUser = { id: lastKey }
const Items = [ { id: getRandomHash() }, lastUser ]
setupMockClient('scan',{
      Items, LastEvaluatedKey: lastUser
    })
return getUsers({ limit: Items.length })
```

backend/__tests__/utils.js
falcoagustin: Good catch 😄 
No specific reason, I was testing it at the time and I forgot to clean it up. A Linter would have proven a lot useful for this cases 😆

backend/common/dynamodb.js
falcoagustin: I always think that it's about the tradeoff, and in this case the style. 
Even though ternaries they are less explicit than to specify if-else blocks, once you get used to them the reading becomes much faster. But there is always the case of what is it that you are doing on each of the blocks, as if it's really big and tied operations that make you loose track that it is happening inside a ternary operation, or you can't notice that it is a ternary operation, then you might wanna go down the explicit road (or refactor it to a new function to fit the one or two line reads of the ternary operation, acting as part of code that smells). 

If it's a simple operation like, do this if true, this if false, I think it's much more readable as you can read the entire logic in one/two lines.

backend/serverless.yml
falcoagustin: Good question 😄 !
As this was a task and I was a bit tight on time, I didn't mind hardcoding some of the values into our configuration. Regardless, for this kind of sensible information I would have used AWS Secrets manager and stored the public key there, so by doing that the fact that it is a public or private key would not matter, as it would have been handled just like any other secret

falcoagustin:
Thanks for taking the time for reviewing my application and for your comments! Great insight.

Regarding logging, I specified much of it on the documentation of the backend `backend/docs/`, I also implemented a middleware that logs each of the operations "locally". Instead, using a tool like [Logstash](https://www.elastic.co/logstash) will bring incredible benefits when watching logs.

For distributed logging, each of the lambdas could receive a header with a unique request Id or generate one if none is specified so that we can have a clear idea, and even query the logs, of what happened and how. For successful or unsuccessful, I'd extend the middleware applier I created by simply:
```js
const result = currentResolver(root, args)
anExamplePostMiddleware(root, args, result)
return result
````
We could also create composable structure (being it more generic, but also more complex) so that we can apply middlewares in a specific order.

Thanks for everything!

File: backend/__tests__/userMutations.test.js [hunk 6]
```diff
@@ -0,0 +1,110 @@
...
+    resetMockClient();
+  });
+
+  test('creates user successfully', () => {
+    setupMockClient('put', {})
+
+    return createUser(DEFAULT_TEST_USER)
+      .then(r => {
+        expect(r).not.toBe(null)
+      })
```
marcelino-sf, line 23: How do you feel about using async/await keywords? Do you think it would make the codebase more succint?

backend/__tests__/userMutations.test.js
```diff
@@ -0,0 +1,110 @@
...
+    expect(() => {
+      updateUser({ ...DEFAULT_TEST_USER, description: '' })
+    }).toThrow(ValidationError)
+  });
+
+  test('update user with non existing id', () => {
+    setupMockClient('update', {}, () => new Error('User does not exists'))
+    expect(() => {
+      updateUser({ ...DEFAULT_TEST_USER, description: '' })
+    }).toThrow(ValidationError)
```
marcelino-sf, line 94: What if the ValidationError passes on this field but not on another, how do you make sure it's validating the correct field?

backend/common/dynamodb.js
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const getClient = () => {
+  return client
+}
```
marcelino-sf, line 11: Do you have a guideline to always use `{}` instead of just `const getClient = () => client`, if so, why would you opt for that?

backend/common/resolverErrors.js
```diff
@@ -0,0 +1,8 @@
+const ValidationError = (message = '') => {
+  this.name = 'ValidationError';
+  this.message = message;
+}
+
+ValidationError.prototype = Error.prototype;
+
+exports.ValidationError = ValidationError
```
marcelino-sf, line 8: Is there a reason you didn't create a class here?

```javascript
class ValidationError extends Error {
  constructor(message = '') {
    super(message);
    this.name = 'ValidationError';
  }
}
```

backend/graphql.js
```diff
@@ -0,0 +1,11 @@
+const { ApolloServer } = require('apollo-server-lambda');
+const { schema: typeDefs } = require('./schema');
+const { resolvers } = require('./resolvers')
+
+const server = new ApolloServer({ typeDefs, resolvers });
+
+exports.graphqlHandler = server.createHandler({
+    cors: {
+        origin: '*',
```
marcelino-sf, line 9: In a real-world scenario, how would you manage this security?

backend/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "jest": {
+    "testPathIgnorePatterns": [
+      "utils"
+    ]
+  },
+  "dependencies": {
+    "apollo-server-lambda": "^2.19.2",
+    "aws-sdk": "^2.839.0",
```
marcelino-sf, line 12: What do you think about adding this `aws-sdk` to devDependencies to solve bundle size? Is it already being managed?

backend/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "graphql": "^15.5.0",
+    "moment": "2.24.0"
```
marcelino-sf, line 14: This library increases the bundle size by 4MB, what do you think about this library, and when it's necessary?

backend/resolvers/index.js
```diff
@@ -0,0 +1,44 @@
+const { userQueries } = require('./users/userQueries');
+const { userMutations } = require('./users/userMutations');
+const { client: { initializeClient }} = require('../common/dynamodb')
+
+// TODO: Improve design by generically applying a middleware engine, and applying this MW to all the resolver functions
+const argumentsMiddleware = (msj, args) => console.log(`${msj}${JSON.stringify(args)}`)
+initializeClient()
```
marcelino-sf, line 7: Considering SOLID principles, do you think the `initializeClient()` should be called here or managed somewhere else?

backend/resolvers/users/userMutations.js
```diff
@@ -0,0 +1,100 @@
...
+
+const createUser = (args) => {
+  const { name, dob, address, description } = args
+  const date = moment(dob, DEFAULT_DATE_FORMAT)
+  if (!date.isValid()) {
+    throw new ValidationError('Invalid date format for field: [dob]; Format should be: ' + DEFAULT_DATE_FORMAT)
+  }
+
+  if (!name || !address || !description) {
+    throw new ValidationError('Invalid empty input for one of the fields: [name, address, description]')
```
marcelino-sf, line 17: Considering the application meant to grow in codebase in a near future, how would you manage the validations?

backend/resolvers/users/userMutations.js
```diff
@@ -0,0 +1,100 @@
...
+    updatedAt: now
+  };
+  const params = {
+    Item: item,
+    TableName: TABLE_NAME
+  };
+
+  return getClient().put(params)
+    .promise()
+    .then(_ => {
```
marcelino-sf, line 38: Can you help me understand why you chose `_` instead of `()`?

backend/resolvers/users/userQueries.js
```diff
@@ -0,0 +1,97 @@
...
+      })
+        .then(r => {
+          return r['features']
+        })
+        .catch(error => {
+          throw new Error('Something bad has happened ' + error)
+        })
+
+    })
+}
```
marcelino-sf, line 92: Can you help me understand why you chose to not create a test-case for this method? If you need to create it, what steps would you do to facilitate the test creation?

backend/schema/index.js
```diff
@@ -0,0 +1,74 @@
+const { gql } = require('apollo-server-lambda');
+
+// TODO: Map all underscore properties from MapBox into camel case for consistency
```
marcelino-sf, line 3: In a real-world project, how do you handle tech-debts to not lose track of them?

marcelino-sf (commented):
Hey @falcoagustin,

Thank you for taking your time to submit this Pull Request!

I left a few comments and would be happy to read your thoughts 😊

Cheers!

backend/__tests__/userQueries.test.js
```diff
@@ -0,0 +1,38 @@
+const { ValidationError } = require('../common/resolverErrors')
+const { getRandomHash } = require('../common/hash');
+const { utils: { setupMockClient, resetMockClient } } = require('./utils');
+const { userQueries: { getUsers }} = require('../resolvers/users/userQueries');
+
+describe('test get user functionality', () => {
+  beforeEach(() => {
+    resetMockClient();
```
pedro-sf, line 8: Have you ever used `jest.spyOn` to mock methods inside other modules?

backend/resolvers/users/userQueries.js
```diff
@@ -0,0 +1,97 @@
...
+        data: result.Items,
+        lastKey: result.LastEvaluatedKey ? result.LastEvaluatedKey['id'] : null
+      }
+    })
+    .catch(error => {
+      throw new Error('Something bad happened: ' + error)
+    })
+};
+
+const getUserLocation = (args) => {
```
pedro-sf, line 43: Why not extract this function to a separate file? :)

backend/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+I understand that the error handling is not that granular, but this is because as most of the errors are validated (for which I provided a `ValidationError` to use), other errors can happen when Dynamo things go wrong.
+When these unexpected things happen, then we could fail fast (as it is a user-intensive application; the end user is the one consuming it) and then send the error into another lambda that takes care about the error handling.
+If the case was with Dynamo, then we should set up an alarm to notify us immediately (as for example when we need more capacity for DynamoDB after pre-provisioning). Other tools can be used for example to call us or send us a slack message like PagerDuty.
+
+In the case of retries, this is error and operation dependent. If we are reading from users and the frontend has already consumed our response, then retrying that operation is completely useless and will only make us pay more for our infra stack, because the user won't notice.
+Given that there is no UI for creating a user, that could be an example of where a retry can happen, and we could add a max of 3 times to retry or then fallback into a DLQ that can retry some time later.
```
pedro-sf, line 28: Let's imagine you built the UI for creating a user:

- Wherein your project would the retry logic live?
- How would you prevent the retry logic to create duplicate records?

pedro-sf (commented):
Hi @falcoagustin, thanks for taking the time to submit this PR 😄 

I have a few questions regarding your code and would love to hear your thoughts 🧠

backend/__tests__/userMutations.test.js
falcoagustin: It'd definitely makes sense to me! I think it makes the code a lot more readable and you can forget about chaining promises, which makes it tidier. This is at the cost of using try/catch blocks to handle errors, but good path in my opinion becomes a lot more easy to read.
The only reason I used promises here instead of using async/await was because of a time constraint and given that the scaffolding project from serverless was already using promises, and started from there 😄

backend/__tests__/userQueries.test.js
falcoagustin: I actually have 😄 
I was having issues while trying to mock `DynamoDB.DocumentClient` (some of them known in github), even tried with [aws-sdk-mock](https://www.npmjs.com/package/aws-sdk-mock) but it didn't work at once and as I was on time constraints, I decided to go with the nice and easy dependency injection approach (injecting a dynamodb mock for tests and actual instance for codebase)

backend/common/dynamodb.js
falcoagustin: Yes I do actually 😄 
I don't use the approach of doing `const myFunc = () => returnValue` for the simple reason that if I need to debug that method with some console or debugger to see what's going on, then I have to write once again the brackets and insert that return logic inside it (to add a simple console statement), do the debugging and clean it up afterwords. That has proven to me really tedious in the past so I decided to do go with the return approach, as IDE helps a lot with this, and it has proven a lot smoother in my opinion

backend/common/resolverErrors.js
falcoagustin: Good one!
No reason at all, I'm used to work with first class functions and I try to be consistent across the entire codebase.

backend/docs/index.md
falcoagustin: Great question 😄 
The retry logic I'd implement it directly with our IaC framework so that we can retry the lambda a maximum amount of times. Given that here the only thing that can happen is that dynamodb fails meaning it didn't insert it correctly for whatever reason, there shouldn't be any risks in duplicating the record as it will retry with the same data. 

This could also be a signal of the platform having to go into the asynchronous design approach, by using an SQS queue to file the creation of a user and having lambdas polling from it to create the user.

backend/graphql.js
falcoagustin: Narrowing it down to the only domain that consumes our graphQL API in this case, or pointing our gateway into s3 so that whenever we are serving static files, they have the same domain and CORS isn't triggered by the browser as the origin is the same.

There are also other approaches when working with large organizations and that is having their own gateway service and making modifications on the fly from a given service to specific registered origins. This can prove troublesome though

backend/package.json
falcoagustin: Good insight 😄 
From my understanding I think that `aws-sdk` is used inside the codebase and not used as a development dependency (as for example `eslint` can be) so I think it should be fine leaving it here.

backend/package.json
falcoagustin: Totally right about that, the size might not serve the purpose for using it for once or twice in the codebase.
I think that library handles dates like no other in a complete and stable way, as it's been used and tested by millions of users. It's really easy to work with dates using that so it's actually my default package to install when working with dates as I'm used to it and don't want to play around creating own stuff (which I have to maintain and test) for shuffling with dates.

On the when is it necessary side, then it's not "necessary" in my opinion, as it just solves a problem in a nicer way than using plain JS or other libraries that make your life easier when working with dates 😄 . If you are on heavy date usage, don't want to maintain and don't have an issue on the bundle size, then that would be a candidate for using momentjs.

backend/resolvers/users/userMutations.js
falcoagustin: Nice one!
I would create a validator that being specified which fields should not be empty it inspects the `args` object and by iterating the keys checks that the values are not empty. This comes with the plus that can be really reused across all codebase and can return a more specific message about which field is broken.

PD: Also relates to the answer on the testing question 😄

backend/resolvers/users/userMutations.js
falcoagustin: Good catch, should have gone with `()`. I used it because the first argument from the resolver is not used for the actual mutations but still passed by

backend/resolvers/users/userQueries.js
falcoagustin: Indeed! 
I thought about that and to actually extract the then part of this function after we get the user from `dynamodb` and getting the location (which can have nothing to do with the concept of user, just the concept of address) so that we have a less coupled codebase. I didn't have much time that weekend to do some improvements, and having a working example was a better choice back then 😄

backend/resolvers/index.js
falcoagustin: That is a great question!
I think that a better way to solve this is to create a service that actually takes care about instantiating this kind of objects. This is where we can use something like IoC frameworks that invert our dependencies and leave the responsibility of creating that instances to another entity.

backend/resolvers/users/userQueries.js
falcoagustin: Sure 😄 
To be honest, I didn't have much time that weekend and wanted to provide a working example with most of the things covered, at least to give an impression of how I worked, and this function was the "longest" one to test (& do the mocking), and I chose to tackle other points on the assignment as I considered them more valuable as there were already several tests to show how would I approach them.

backend/schema/index.js
falcoagustin: Great question!
TODOS + tickets is the way to go. Leaving TODOs is a must I think so that it breaks your eyes when reading the codebase, making you remember that you have something to work on.
Creating tickets is for the managing side so that we can make room in the sprint for this upcoming improvements.

falcoagustin (commented):
@marcelino-sf and @pedro-sf Thanks for taking the time to review and comment!

Some of the decisions I took where based on the time constraints I had, deciding to provide a more complete solution but still showing always the way of how things should or could be approached.

Thank you for everything!

PR #36, opened 2021-03-08
Backend Challenge Solution

Manghud:
Here is my solution to the backend challenge. If you want to play around the live demo, please don't hesitate to ask me for an API key


Manghud:
> Here is my solution to the backend challenge. If you want to play around the live demo, please don't hesitate to ask me for an API key

Forgot to add that a description of the solution and some of my reasoning can be found on the branch's updated README

resolvers/crud/create/response.vtl
```diff
@@ -0,0 +1 @@
+$util.toJson($context.result)
```
patrick-sf, line 1: How would you handle the situation where **$context** contains an error property?

resolvers/crud/delete/request.vtl
```diff
@@ -0,0 +1,7 @@
+{
```
patrick-sf, line 1: What is the expected behavior if the specified **id** to delete does not exist in the table?

services/getLocationInfo/src/index.ts
```diff
@@ -0,0 +1,22 @@
...
+export const LocationService:LocationAPI = new MapBoxAPI();
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const handler = async (event:AppSyncEvent, _context?:any) => {
+  if (!event.arguments) {
+    return [];
+  }
+  const { address } = event.arguments;
+  try {
+    const locations = await LocationService.getLocationData(address);
```
patrick-sf, line 15: If you were to add a validation step, how would you enforce proper address formatting to allow for a successful query?

README.md
```diff
@@ -1,130 +1,82 @@
...
+- GetLocationInfo Query supported by a AWS Lambda function with `npm` dependencies bundled as AWS Lambda Layers
+
+## Error Handling Strategy
+At the moment, the provided strategy is a simple logging mechanism that stores formatted error objects to CloudWatch. However, this can be improved in a variety of ways:
+- My first improvement would be adding a DLQ for each Lambda function, in order to use it as a target buffer for failed Lambda Messages. After setting up the DLQ, we could build another lambda to consume this messages and reprocess them if necessary. For the given requirement of returning Location Data, requeuing failed attempts does not seem fitting.
+- Another improvement would be adding a CloudWatch Alarm and a Notification system (using SNS and another Lambda function) in order to alert the respective support team/devs via email or sms.
+
+## Wishlist
+I must admit that this solution was built with very little Terraform knowledge, so I'm sure there is lots of room for improvement in terms of making Terraform code more modular, having reusable resources, and a better way to set parameters (thinking of CI/CD integration).
```
patrick-sf, line 81: Was this effort (Terraform) enjoyable or unpleasant, and why?

patrick-sf (commented):
Hi! Thanks for taking the time to work on this challenge/solution.

I am looking forward to getting to know your thought process and have left some questions for you.

.gitignore
```diff
@@ -0,0 +1,8 @@
+.terraform/**
+terraform.tfstate*
+variables.tf
```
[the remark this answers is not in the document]
Manghud: I've researched a bit more about this topic. Unlike my current implementation, the way to go seems to keep the variables.tf file with default or dummy values in the repo and moving the actual values to a `tfvars` file which should be kept out of source control. The `tfvars` file should then be securely stored as CI/CD credentials, or whichever secret vault that is being used in the project.

api.tf
```diff
@@ -0,0 +1,116 @@
...
+  authentication_type = "API_KEY"
+  xray_enabled = var.enableTracing
+  log_config {
+    cloudwatch_logs_role_arn = aws_iam_role.iam_appsync_role.arn
+    field_log_level = var.apiLoggingLevel
+  }
+}
+
+# Create the API key.
+resource "aws_appsync_api_key" "appsync_api_key" {
```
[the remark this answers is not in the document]
Manghud: I assume you mean looking at the API key value as a developer owning the repo and having synchronized with the securely deployed state file. 
The way to go would be to define an output for this variable and set the `sensitive` flag to true. This is necessary because I'm assuming that we would surely refactor this into a separate module and thus would possibly have to reference this value from outside modules. The `sensitive` flag would prevent the key to be printed in stdout or logfiles. In order to read the sensitive data you would have to parse or `grep` the respective state file that should also be secured properly.

api.tf
```diff
@@ -0,0 +1,116 @@
+# IAM role
```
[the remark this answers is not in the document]
Manghud: I think the right approach would be to delegate specific policy mappings and dynamoDB resolvers to a generic child module that abstracts those settings.
If we combine that approach with input variables (e.g. to set the id of the API for the datasource, table name, and declare a subset of CRUD endpoints to be created), we could have a way to define generic CRUD modules and avoid rewriting most of the configuration for each db table we want to define. A similar approach could be used for the lambda function albeit with a bit less parameterization.

services/getLocationInfo/src/index.ts
Manghud: I think that error handling logic should be improved in a way that enables throwing specific error types for different scenarios. Then, instead of throwing a standard error, I would simply return the error object (let's say some `RequestValidationError` object). This approach would rely on having clearly defined error types in the GraphQL schema and returning a Union Type from the query that could be either the expected data type or as many possible error types for that specific query. 
In order to actually validate if the address follows the expected format, I would move that logic to a separate class whose sole responsibility would be to validate event arguments. In there, we should first try to sanitize the string and possibly test it against a Regexp that fits the given requirements.

resolvers/crud/create/response.vtl
Manghud: According to the official docs: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html 
That scenario should not be a real concern given that everything that might have gone wrong with the actual resolver should be included in `$context.result`.
But assuming we would do so, I would rely on checking the existence of that particular property and appending it to `$context.result` as a `contextError` so as to not interfere with any internal error that could be defined within  `$context.result`. I'm just learning VTL but after reading the reference here:  https://velocity.apache.org/engine/1.7/user-guide.html#if-elseif-else I would replace that snippet to something along the lines of:
```
#if(!$context.result)
  #set($context.result = {})
#end
#if( $context.error)
  #set($context.result.contextError = $context.error)
#end
$util.toJson($context.result)
```

resolvers/crud/delete/request.vtl
Manghud: By default, the DeleteItem operation is idempotent, and as such, there should be no error if we attempt to delete a record that does not exist on the table. The expected result should be a successful query with no data

README.md
Manghud: I would say that the learning curve was pretty manageable considering that I was able to learn some of the essentials and build this basic solution in a matter of hours. The rewards of using Terraform are pretty apparent without diving too deep in the benefits of using a infrastructure-as-code solution. I benefited greatly from using both the `validate` and `plan` commands, without those I would have messed up my progress several times over. I also found that the docs are very clear and straightforward, once I found the AWS provider docs I barely had to look elsewhere in order to configure my AWS resources. Definitely something I will keep using for future projects.

getLocationInfo.tf
```diff
@@ -0,0 +1,34 @@
...
+  role          = aws_iam_role.iam_lambda_role.arn
+  filename         = data.archive_file.getLocationInfo_lambda_zip.output_path
+  source_code_hash = data.archive_file.getLocationInfo_lambda_zip.output_base64sha256
+  runtime       = "nodejs12.x"
+  handler = "index.handler"
+  layers = [aws_lambda_layer_version.nodeFetch_lambda_layer.arn]
+  environment {
+    variables = {
+      "MAPBOX_ACCESS_TOKEN" = var.MAPBOX_ACCESS_TOKEN
+    }
```
marcelino-sf, line 20: How would you manage secret variables (in case you have it) on lambda functions?

resolvers/crud/list/request.vtl
```diff
@@ -0,0 +1,7 @@
+{
+  "version": "2017-02-28",
+  "operation": "Scan",
+  "filter": #if($context.args.filter) $util.transform.toDynamoDBFilterExpression($ctx.args.filter) #else null #end,
+  "limit": $util.defaultIfNull($ctx.args.limit, 20),
+  "nextToken": $util.toJson($util.defaultIfNullOrEmpty($ctx.args.nextToken, null)),
+}
```
marcelino-sf, line 7: How would you do in a real-world project that is supposed to scale, do you consider this approach production-ready for large projects?

resolvers/crud/update/request.vtl
```diff
@@ -0,0 +1,83 @@
...
+  "key": {
+    "id": $util.dynamodb.toDynamoDBJson($ctx.args.input.id),
+  },
+
+  ## Set up some space to keep track of things we're updating **
+  #set( $expNames  = {} )
+  #set( $expValues = {} )
+  #set( $expSet = {} )
+  #set( $expAdd = {} )
+  #set( $expRemove = [] )
```
marcelino-sf, line 13: What do you feel about writing code for AppSync? I'm curious to see your thoughts here 😊

services/getLocationInfo/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "test:noWatch": "NODE_ENV=test && jest",
+    "test": "NODE_ENV=test jest --watchAll",
+    "coverage": "NODE_ENV=test jest --coverage"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {},
+  "devDependencies": {
+    "ts-node": "^8.3.0",
+    "@types/node-fetch": "^2.5.8",
```
marcelino-sf, line 20: Can you help me understand what were your reasons to pick `node-fetch` instead of something else?

marcelino-sf (commented):
Hey @Manghud, thank you very much for taking your time creating this Pull Request!
I left a few questions, looking forward to hearing your thoughts on them! 😊

Cheers! ✌️

services/getLocationInfo/src/index.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { mocked } from 'ts-jest/utils';
+
+import { handler, LocationService } from './index';
+import { AppSyncEvent } from './event';
```
pedro-sf, line 4: There is a nice package called `@types/aws-lambda` that handles typing most of AWS's events for you :)

pedro-sf (commented):
Thank you for taking the time to submit this pull request! 😄 

I've added some comments and am keen on hearing your thoughts. 🧠

getLocationInfo.tf
Manghud: Hello @marcelino-sf.
By default, AWS Lambda encrypts environment variables with a default key that is AWS managed, the downside is that the variables could still be seen if access to the AWS console UI is compromised. Alternatively, we could setup a way to use our own managed master key (CMK) in order to protect environment variables.
However, the more secure way of doing this would be to encrypt the environment variable from the client side first (I would attempt to do so on Terraform using the `kms_key_arn ` parameter) and then attempt to decrypt it from within the Lambda's code using AWS KMS from within the lambda function during runtime.

resolvers/crud/list/request.vtl
Manghud: I would not say this is production ready. First of all, I would move this logic to it's own Lambda resolver instead of relying on VTL resolvers. This would allow to have more readable and testable implementation. Even though the `$util.transform.toDynamoDBFilterExpression` helper should probably do parameter sanitization internally, I would still choose to add my own logic for that as an additional layer.

Manghud: Also, considering that we are discussing a large project (in terms of data). I would be very careful with which queries are allowed to perform scans like this one in order to prevent overloading the database. In contrast with this basic implementation, every query should be backed by a database index.

resolvers/crud/update/request.vtl
Manghud: I found that getting a basic GraphQL CRUD API up and running is amazingly simple. The auto generated schema even defines subscriptions that allow clients to easily get real time data. I'm not a big fan of VTL but fortunately we could implement direct Lambda resolvers and have our custom-made logic. I feel like this service allows us to focus more on building business logic by saving us all the trouble of configuring and managing a GraphQL server by ourselves.
My only previous experience with GraphQL was  a few years ago (2018) I remember having to implement data loaders (n+1 problem) manually, there were no well established patterns on how to do authentication/authorization and overall it seemed like it was a very immature ecosystem. So I'm a bit in awe at how much has all of it evolved from that point. I will surely consider using this service whenever I look to expose a GraphQL on AWS

services/getLocationInfo/package.json
Manghud: Sure thing. As you imply, there are quite a few alternatives for this, including node's default `http` module. First of all, I would want to avoid having to "promisify" or having to care about the abstractions and gotchas that most npm modules already handle.
Having said that, the more widely used and community-backed npm modules I would consider are [node-fetch](https://www.npmjs.com/package/node-fetch), [superagent](https://www.npmjs.com/package/superagent), and [axios](https://www.npmjs.com/package/axios). Out of those, node-fetch is the most lightweight, the second most popular, the one with least amount of unresolved github issues, is very well documented, and has the added benefit of mirroring the API of the browser's fetch library.

services/getLocationInfo/src/index.spec.ts
Manghud: Hi Pedro, that's neat. I've mostly worked with SQS events but didn't know about that library. I will definitely check that out, thanks!

README.md
```diff
@@ -1,130 +1,82 @@
...
+- **Data Store**: DynamoDB
+- **Custom Location Service**: NodeJS (Typescript) + AWS Lambda
+- **IAC**: Terraform
+- **Location API**: Mapbox
+
+## General Description and How to Run
+The provided solution uses Terraform in order to provision AWS infrastructure. In order to deploy the project, a `variables.tf` file must be defined on the root of the project in order to define necessary parameters such as AWS credentials and MapBox access token. Here's a sample file:
+
+
+    variable "aws_credentials_profile" {
```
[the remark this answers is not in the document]
Manghud: I think I mentioned some of this on a comment made by @patrick-sf 
> I've researched a bit more about this topic. Unlike my current implementation, the way to go seems to keep the variables.tf file with default or dummy values in the repo and moving the actual values to a `tfvars` file which should be kept out of source control. The `tfvars` file should then be securely stored as CI/CD credentials, environment variables, or whichever secret vault that is being used in the project.

getLocationInfo.tf
```diff
@@ -0,0 +1,34 @@
...
+
+# Create lambda function from zip file, with lambda role.
+resource "aws_lambda_function" "getLocationInfo_lambda" {
+  function_name = "${var.prefix}_getLocationInfo_lambda"
+  role          = aws_iam_role.iam_lambda_role.arn
+  filename         = data.archive_file.getLocationInfo_lambda_zip.output_path
+  source_code_hash = data.archive_file.getLocationInfo_lambda_zip.output_base64sha256
+  runtime       = "nodejs12.x"
+  handler = "index.handler"
+  layers = [aws_lambda_layer_version.nodeFetch_lambda_layer.arn]
```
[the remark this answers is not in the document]
Manghud: For this small project I could not find any drawback apart from having to remember to do an `npm install` within the layer's directory before deploying. Having dependencies as layers allows to have a clean and lightweight zip file deployed instead of uploading `node_modules` alongside the function. Smaller bundle size also means that cold starts are somewhat mitigated.

For a larger project I would have to be aware that other AWS lambdas might also depend on this layer. And finally, thinking on the mid to long term, I would have to keep track of the layer version for the inevitable module upgrade. For instance, if a security patch was released and thus would have to manually do an upgrade for the layer's code and deploy the new version.

README.md
```diff
@@ -1,130 +1,82 @@
...
-- New Jersey
-- Northern Bergen County, NJ
-
-> I almost finished, but I don't have time to create everything what is required
-
-Please provide a plan for the rest of the things that you would do.
+This is the solution for the Backend section of the test.
+
+### Live Demo
+Use GraphQL api [here](https://pckwbzav6nazdnitbphembx2ne.appsync-api.us-east-2.amazonaws.com/graphql/ "Live API Demo"). An API key can be provided by sending me an email request.
```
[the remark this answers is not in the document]
Manghud: I would consider an API Key as a secret that not even the client code should access directly (hard coding the key on an app's code would be giving it away pretty much), nevermind public information. The only scenario I could think of where an API key could be public information would be to restrict access to a very limited sandbox or demo-site (I didn't do so for this challenge, ironically :sweat_smile:).

PR #39, opened 2021-06-13
Anderson Leal web submission

andersonleal1805:


.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
```
marcin-piela-sf, line 3: Is it your preferred prettier config? Do you use anything more in a real project?

README.md
```diff
@@ -1,9 +1,41 @@
...
+- This command will update the schema.graphql file and generate the hooks for each query defined on the folder.
+
+
+- To start the project
+    ```
+        $ npm run start
+    ```
+
+
+- [Working Demo](https://superformula-c6aa8.web.app/)
```
marcin-piela-sf, line 30: It's not working cause this version is connected with `localhost` API :) It's not a problem I configured it on my local machine

package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "test-full-stack",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@apollo/client": "^3.3.19",
+    "@testing-library/jest-dom": "^5.11.4",
+    "@testing-library/react": "^11.1.0",
+    "@testing-library/user-event": "^12.1.10",
+    "@types/jest": "^26.0.15",
```
marcin-piela-sf, line 10: What are your thoughts about moving it to `devDependencies`?

src/App.tsx
```diff
@@ -0,0 +1,22 @@
...
+          <UserEditModalProvider modalComponent={<UserEditModal />}>
+            <UserListPage />
+          </UserEditModalProvider>
+        </QueryParamProvider>
+      </Router>
+    </ApolloClientProvider>
+  );
+}
+
+export default App;
```
marcin-piela-sf, line 22: What are the disadvantages of using default export in React.JS applications

src/App.tsx
```diff
@@ -0,0 +1,22 @@
+import { ApolloClientProvider } from './commons/ApolloClientProvider/ApolloClientProvider';
+import { UserListPage } from './pages/UserList';
+import { UserEditModalProvider } from './components/modal/ModalComponent';
+import { QueryParamProvider } from 'use-query-params';
+import { BrowserRouter as Router, Route } from 'react-router-dom';
```
marcin-piela-sf, line 5: Do you know any way to order `import statements` automatically? Let's say I want to have `external` imports first

src/components/MapBoxStaticMap/MapBoxStaticMap.module.scss
```diff
@@ -0,0 +1,7 @@
+.element {
+  border-radius: 8px;
```
marcin-piela-sf, line 2: How would you share such values between components in a real project?

src/components/MapBoxStaticMap/MapBoxStaticMap.tsx
```diff
@@ -0,0 +1,19 @@
...
+    <div
+      className={classes.element}
+      style={{
+        backgroundImage: `url("${props.src}")`,
+      }}
+    />
+  );
+}
+
+export const MapBoxStaticMap = memo(MapBoxStaticMapComponent);
```
marcin-piela-sf, line 19: What is the purpose of using `memo` in all components?

src/components/avatar/Avatar.module.scss
```diff
@@ -0,0 +1,7 @@
+.element {
```
marcin-piela-sf, line 1: What would you say about renaming all these `element` CSS classes to name of the component like:
```
.avatar {
```

src/components/avatar/Avatar.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import classes from './Avatar.module.scss';
+
+interface AvatarProps {
+  src: string;
+  title: string;
+  className?: string;
+}
+
+function AvatarComponent(props: AvatarProps) {
```
marcin-piela-sf, line 12: What are your thoughts on destructuring component props, like 
```
{className}: AvatarProps
```
?

src/components/button/Button.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  border-radius: 8px;
+  height: 90px;
+  padding: 11px 14px;
+  cursor: pointer;
+
+  font-weight: 600;
+  font-size: 24px;
+  line-height: 30px;
+
+  &.secondary {
```
marcin-piela-sf, line 14: What about using CSS methodology like BEM here?
```
avatar--secondary
```

src/components/button/Button.tsx
```diff
@@ -0,0 +1,29 @@
+import { memo, MouseEventHandler } from 'react';
+import { ChildrenProps } from '../General';
+import classnames from 'classnames';
+
+import classes from './Button.module.scss';
+
+interface ButtonProps extends ChildrenProps {
```
marcin-piela-sf, line 7: Looks like this interface is incorrect cause `children` should be required here but it is optional in `ChildrenProps`. Do you prefer to have a global `ChildrenProps` type in the project?

src/pages/UserEdit/UserEditModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+  onSubmit: () => void;
+  latitude?: number;
+  longitude?: number;
+  onInputChange: (name: string) => ChangeEventHandler;
+}
+
+function UserEditModalComponent(props: UserEditModalProps) {
+  return (
+    <div className={classes.element}>
+      <h1>Edit user</h1>
```
marcin-piela-sf, line 21: What library would you recommend to handle i18n in the real app?

src/pages/UserEdit/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+          variables: {
+            address: form.address,
+          },
+        });
+      }, 600);
+    }
+  }, [form, fetchAddress]);
+
+  const onInputChange = useCallback(
+    (name: string) => (evt: React.ChangeEvent<HTMLInputElement>) => {
```
marcin-piela-sf, line 46: would you recommend any library to handle this `form` logic?

src/pages/UserList/UserCard/UserCard.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <Card className={classes.element}>
+      <img
+        src="assets/edit-icon.svg"
+        alt="Edit icon"
+        className={classes.editIcon}
+        onClick={props.onEditClick}
+      />
+      <Avatar
+        title={'avatar'}
```
marcin-piela-sf, line 24: Sometimes you write:
```
stringProp="string"
```
and sometimes:
```
stringProp={'string'}
```
What is your preferred way? How to make it consistent across the project?

src/pages/UserList/index.tsx
```diff
@@ -0,0 +1,92 @@
+import { memo, useCallback, useEffect, useRef, useState } from 'react';
+import { UserListPage as Element } from './UserListPage';
+import { useFetchUsersLazyQuery, User } from '../../generated/graphql';
+import { useUserEditModal } from '../../components/modal/ModalComponent';
+import { StringParam, useQueryParam } from 'use-query-params';
+
+function UserListPageComponent() {
+  const [fetch, { data, fetchMore, loading }] = useFetchUsersLazyQuery();
```
marcin-piela-sf, line 8: What happens when this query fails?

src/components/modal/ModalComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+  modalComponent: ReactNode;
+}
+
+function UserEditModalProviderComponent(props: ModalProviderProps) {
+  const [user, setUser] = useState<User>();
+  const [isShow, setShow] = useState(false);
+  const { children, className } = props;
+
+  const openDialog = useCallback((user: User) => {
+    setUser(user);
```
marcin-piela-sf, line 43: Why do we have such`logic` inside the generic `Modal` component?

src/components/modal/ModalComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <div className={classNames(classes.modal, className)}>
+        <div className={classes.backdrop} onClick={closeDialog} />
+        <div className={classes.modalWindow}>{props.modalComponent}</div>
+      </div>,
+      document.body
+    );
+  }, [className, closeDialog, isShow, props.modalComponent]);
+
+  return (
+    <ModalContext.Provider value={controller}>
```
marcin-piela-sf, line 71: What are the advantages of using `Context` inside `Modal` instead of simply passing `isOpen`, `onClose`, `onOpen` props to this component?

marcin-piela-sf (commented):
Thanks, @andersonleal1805  for your submission :) Good job 👍 

README.md
andersonleal1805: ohh man 🤦 , I did a build with localhost, did a new build, and deployed the version. Now it works.

.prettierrc
andersonleal1805: I don't know, for this project I preferred to keep it simple. What would be a good option to add to this config?

package.json
andersonleal1805: yes, I agree, that should be in dev dependencies.

src/App.tsx
andersonleal1805: we could use prettier for that, like [prettier-plugin-import-sort](https://www.npmjs.com/package/prettier-plugin-import-sort) library

src/App.tsx
andersonleal1805: With export default, we can get an inconsistent Component name. Like `import MyAwesomeApp from './App';`

src/components/MapBoxStaticMap/MapBoxStaticMap.module.scss
andersonleal1805: I can create a `common.scss` and define all variable there and import and use on the scss files

src/components/MapBoxStaticMap/MapBoxStaticMap.tsx
andersonleal1805: Prevent the component from making a render unnecessarily, which can result in a loss of application performance.

src/components/avatar/Avatar.module.scss
andersonleal1805: I liked the suggestion, I can do that and start using the component name.

src/components/avatar/Avatar.tsx
andersonleal1805: I like to use variable destructuring, but there are cases where the component has too many props leaving the function signature too big. But yes, in this case, I could have used destructuring.

src/components/button/Button.module.scss
andersonleal1805: Yes, BEM is a good strategy to keep the module well defined. I will change the project to follow this strategy.

src/components/button/Button.tsx
andersonleal1805: You're right, the `children` on `ChildrenProps` interface shouldn't be optional. I was thinking about making the components more consistent with `ChildrenProps`

src/components/modal/ModalComponent.tsx
andersonleal1805: This was a forgotten logic 🤦 , initially this modal was only for user editing, then I thought why not make it a generic modal, so this rule ended up getting lost there. I'll fix it.

src/components/modal/ModalComponent.tsx
andersonleal1805: I did it that way to show an implementation with `React.context`. I agree that it would be much simpler to pass `isOpen` to Modal component. 😄

src/pages/UserEdit/UserEditModal.tsx
andersonleal1805: I've already used [react-intl](https://formatjs.io/docs/react-intl/), very simple to use.

src/pages/UserEdit/index.tsx
andersonleal1805: I usually use [react-final-form](https://final-form.org/react) in my projects. I decided not to add to this project as it is a very simple form with 3 inputs.

src/pages/UserList/UserCard/UserCard.tsx
andersonleal1805: I prefer single quotes, I added this rule to prettier 😃 . `"jsxSingleQuote": true`

src/pages/UserList/index.tsx
andersonleal1805: In this case, I'm not displaying anything 🤦. I'm not using the error object, I'll fix it.

PR #40, opened 2021-06-21
Marcelo Aleixo

marcelo-aleixo:
Please check **README_I.md**, I didn't want to overwrite README.md (The challenge info)

Working demo: https://marcelo-aleixo-sf-fe.web.app


src/components/input/Input.tsx
```diff
@@ -0,0 +1,33 @@
...
+}) => (
+  <div className={classnames('input-wrapper', className)}>
+    <label htmlFor={name}>{label}</label>
+    <input
+      className="input-field"
+      name={name}
+      placeholder={placeholder}
+      onChange={onChange}
+      value={value}
+      tabIndex={0}
```
mat-sf, line 30: Why we set it to 0 here? What do you think about in general changing tabIndex'es?

src/components/input/Input.scss
```diff
@@ -0,0 +1,45 @@
...
+      display: block;
+      font-weight: 600;
+      font-size: 18px;
+      line-height: 23px;
+      color: $color-black;
+      margin-bottom: 8px;
+    }
+  }
+
+  &-field {
```
mat-sf, line 20: Do you write class names with any methodology? If not, do you have any experience in using any?

src/components/input/Input.scss
```diff
@@ -0,0 +1,45 @@
...
+  &-wrapper {
+    padding: 0;
+    display: inline-block;
+    width: 400px;
+    max-width: 100%;
+
+    label {
+      display: block;
+      font-weight: 600;
+      font-size: 18px;
```
mat-sf, line 13: WDYT about using px for font-sizes?

src/components/input/Input.scss
```diff
@@ -0,0 +1,45 @@
...
+      line-height: 23px;
```
mat-sf, line 14: WDYT about using pixels for `line-height`s?

src/components/modal/DefaultModalFooter/__tests__/DefaultModalFooter.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    render(
+      <DefaultModalFooter
+        onCancelClick={onCancelClick}
+        onCancelText={onCancelText}
+        onOkClick={() => {}}
+      />,
+    );
+
+    act(() => {
```
mat-sf, line 23: Can you elaborate more on why do we have to wrap those events in `act`?

src/components/typograph/Typography.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface CardProps {
+  variant?: VariantEnum
+  color?: ColorEnum
+  className?: string
+}
+
+export const Typography: React.FC<CardProps> = ({
+  children, variant, className, color = 'primary',
+}) => (
+  <div className={classnames(className, {
```
mat-sf, line 17: Can you elaborate more on why do you prefer to use `div`s instead of proper text tags?

src/hooks/useUpdateUser.ts
```diff
@@ -0,0 +1,32 @@
...
+interface GeocodingInt {
+  geocoding: Geocoding
+}
+
+const useUpdateUser = () => {
+  const [updateUser, {
+    data, error, loading,
+  }] = useMutation<GeocodingInt>(UPDATE_USER);
+
+  const update = useCallback((user: UpdateUserInput) => updateUser({
```
mat-sf, line 15: Why do we wrap it in `useCallback`?

src/Views/UsersPage/UsersPage.tsx
```diff
@@ -0,0 +1,69 @@
+import React, { useCallback, useEffect } from 'react';
+import { debounce } from 'lodash';
```
mat-sf, line 2: What do you think about making direct imports from lodash?

src/Views/UsersPage/UsersPage.tsx
```diff
@@ -0,0 +1,69 @@
...
+          />
+        </div>
+      </>
+      )}
+
+    </div>
+  );
+};
+
+export default Users;
```
mat-sf, line 69: What do you think about named exports vs default exports, any advantages/disadvantages of one or the other solution?

src/Views/UsersPage/UserList/UserList.tsx
```diff
@@ -0,0 +1,18 @@
...
+import './UserList.scss';
+
+interface UserListProps {
+  users?: User[]
+}
+
+const UserList: React.FC<UserListProps> = ({ users }) => (
+  <div className="user-list">
+    {users?.map((user) => (
+      <UserCard key={user.id} user={user} />
```
mat-sf, line 13: If the user would not have `id` field, would you use `index` as a key here? If yes/no - why?

.env
```diff
@@ -0,0 +1,4 @@
+SKIP_PREFLIGHT_CHECK=true
+
+# This should not be here, but I left so you can run the application locally and point to my GraphQL Server
+REACT_APP_GRAPHQL_ENDPOINT="https://bjyx82su0i.execute-api.us-east-1.amazonaws.com/dev/graphql"
```
mat-sf, line 4: 👍

src/Views/UsersPage/UserList/__tests__/UserList.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+    description: 'Dolor Amet',
+    id: '5f3c1362-48d9-4ff3-be8b-6544eb26hg4a',
+    imageUrl: 'https://unsplash.com/photos/YWpWKO9PNns/download?force=true&w=640',
+    name: 'John Doe',
+    __typename: 'User',
+  } as User,
+];
+
+describe('<UserList />', () => {
+  it('Should render correctly', () => {
```
mat-sf, line 30: What do you think about having 100% unit tests coverage?

mat-sf (commented):
Hi Marcelo! Thanks for the PR and the time you have spent on this.
I left a few comments so we can have a little discussion, if you have any questions - feel free to ask 🙂

src/App.test.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { MockedProvider } from '@apollo/client/testing';
+import { render, screen } from '@testing-library/react';
+import App from './App';
+
+test('renders learn react link', () => {
+  render(
+    <MockedProvider mocks={[]} addTypename={false}>
```
marcin-piela-sf, line 8: Let's imagine that we have an `Auth` global context in our app and we use it in a lot of places, how would you write/configure tests to not wrap the component in every test case?

src/Views/UsersPage/UserList/UserCard/UserCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { User } from '../../../../types';
+import { formatDate } from '../../../../utils/dates';
+import './UserCard.scss';
+import UserEdit from './UserEdit/UserEdit';
+
+interface UserCardProps {
+  user: User
+}
+
+const UserCard: React.FC<UserCardProps> = ({ user }) => (
```
marcin-piela-sf, line 15: What are the disadvantages of using `React.FC`?

src/Views/UsersPage/UserList/UserCard/UserCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </Typography>
+    </div>
+    <div className="user-card-description">
+      <Typography variant="paragraph">{user.description}</Typography>
+    </div>
+    <UserEdit user={user} className="user-card-edit-button" />
+  </Card>
+);
+
+export default React.memo(UserCard);
```
marcin-piela-sf, line 35: In this particular example, will `React.memo` improve app performance?

src/Views/UsersPage/UserList/UserCard/UserCard.scss
```diff
@@ -0,0 +1,67 @@
...
+    display: flex;
+    justify-content: center;
+    margin-bottom: 28px;
+  }
+
+  &-info {
+    width: 100%;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 8px;
```
marcin-piela-sf, line 19: What do you think about moving this value to `variables`?

src/Views/UsersPage/UserList/UserCard/UserEdit/UserEdit.tsx
```diff
@@ -0,0 +1,75 @@
...
+    await update(updatedUser);
+    close();
+  }, [close, update, updatedUser]);
+
+  return (
+    <div className={classnames(className)}>
+      <button type="button" onClick={open} className="user-edit-icon" tabIndex={-1}>
+        <img src={editIcon} alt="edit" />
+      </button>
+      <Modal isOpen={isOpen} title="Edit user" footer={(<DefaultModalFooter onCancelClick={close} onOkClick={submitForm} buttonsDisabled={loading} />)}>
```
marcin-piela-sf, line 40: What is your preferred way?

1. `<Modal footer={}/>`
2. `<Modal>....<Modal.Footer>...</Modal.Footer></Modal>`
3. Something else?

src/__snapshots__/storybook.test.ts.snap
```diff
@@ -0,0 +1,130 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
marcin-piela-sf, line 1: When do you prefer to use `snapshots`?

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface ModalProps {
+  isOpen: boolean
+  title: string
+  footer?: ReactNode
+}
+
+export const Modal: React.FC<ModalProps> = ({
+  children, isOpen, title, footer,
+}) => (isOpen ? ReactDOM.createPortal(
+  <div className={classnames('modal-overlay', { 'modal-overlay-visible': isOpen })}>
```
marcin-piela-sf, line 16: Right now the user is able to scroll the whole page when the modal is opened. How would you fix it?

src/hooks/useScrollPosition.ts
```diff
@@ -0,0 +1,28 @@
+import { debounce } from 'lodash';
+import { useCallback } from 'react';
+
+const useScrollPosition = () => {
+  const scrollPosition = sessionStorage.getItem('scrollPosition');
```
marcin-piela-sf, line 5: Do you know any case when we shouldn't use `sessionStorage`?

src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface ModalProps {
+  isOpen: boolean
+  title: string
+  footer?: ReactNode
+}
+
+export const Modal: React.FC<ModalProps> = ({
+  children, isOpen, title, footer,
+}) => (isOpen ? ReactDOM.createPortal(
```
marcin-piela-sf, line 15: Let's say that we want to animate modal when we close it, how would you achieve that with `createPortal`?

src/GraphQL/mutations/UpdateUser.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from '@apollo/client';
+
+const UPDATE_USER = gql`
+    mutation UpdateUser(
```
marcin-piela-sf, line 4: in `updateUser` mutation we have one `input` and here we're providing 4 different parameters, can you elaborate more on this approach?

marcin-piela-sf (commented):
Thanks for your submission! I left some questions, looking forward to your response :)

src/components/input/Input.tsx
marcelo-aleixo: I was adding the tabIndex -1 to the edit button and for some reason that I don't recall right now I put the tabIndex 0 to the Input element, but it not necessary indeed.

I would not recommend messing with tabIndex, since it can be messy real quick, let's imagine I put positive numbers to all components, then I reorder the layout of the site, what would happen? The tab order would be messy and I would have to re-order the tabIndexes, this is not good at all. 

My eslint configuration catches this and only allows -1 and 0 so this would not mess with the order.

src/components/input/Input.scss
marcelo-aleixo: I used to use BEM, then I migrate to CSS in JS and it was a completely different naming convention, for this project I kept simple and used only: block-element-variant with only one hyphen instead of __ and --.

I even used the variant(--) for the typography component.

src/components/input/Input.scss
marcelo-aleixo: You got me right here, I usually stick to the pattern the project is using, I know em and rem is preferable due to its ability of scale, but I kept it simple in this project and sticked with px.

src/components/input/Input.scss
marcelo-aleixo: Please look the answer above

src/components/modal/DefaultModalFooter/__tests__/DefaultModalFooter.test.tsx
marcelo-aleixo: I use it whenever the action needs to trigger a react state update for that specific component I am testing, like I did for `useToggleStatus.test.ts`, for the scenario above, it wasn't really that necessary since it is triggering an action that does not update the internal react state.

src/components/typograph/Typography.tsx
marcelo-aleixo: 🤦‍♂️ It was actually a mistake that I made from not mapping the variant to its corresponding text tag.

I could have created an object with all possible text tags, then used object literal to retrieve the correct one like so:

```js
const variantMapping = {
  h1: 'h1',
  h2: 'h2',
  paragraph: 'p',
};
```

then used like:

```js
const Component: any = variantMapping[variant];

  return (
    <Component className={classnames(className, {
      [`typography--variant-${variant}`]: variant,
      [`typography--color-${color}`]: color,
    })}
    >
      {children}
    </Component>
  );
```

src/hooks/useUpdateUser.ts
marcelo-aleixo: I use useCallback since it returns a memoized function that is only re-created when one of the dependency changes, this is specially used when passing a function as props since it will not be recreated on each render.

src/Views/UsersPage/UsersPage.tsx
marcelo-aleixo: yeah... Lodash isn't light for FE therefore we have the correct way of importing it.

The way I am doing using the destructuring actually does not differ at all from importing it like `import _ from 'lodash'` regarding tree shaking, it will import the whole Lodash package into the bundle.

I could've used `import 'lodash/debounce'` but my habit made me use the destructuring.

I've worked in a project the I had to do some bundle shrinking and I used webpack-bundle-analyzer to analyze the bundle result, it was scaring to see how lodash was impacting the bundle size. 

I was an ejected project, so I installed babel-plugin-lodash that does that `import 'lodash/...'` automatically.

src/Views/UsersPage/UsersPage.tsx
marcelo-aleixo: That is funny, 'cause I started the project with the ESLint rule`import/prefer-default-export` turned on, but then later on I disabled it.

The good thing about default export is that we can name it whatever we like when importing like in the example above:

```js
import AnotherNameHere from 'Users';
```

But this is easily done with named export too, but we write more:

```js
import { Users as AnotherNameHere } from 'users';
```

The con in default export is that we can only export one thing, or an object.

For eg.

```js
const functionOne = () => {}
const functionTwo = () => {}

export default {
    functionOne,
    functionTwo
}
```

# On the other hand...

I like to export multiple functions with name export:

```js
export const functionOne = () => {}
export const functionTwo = () => {}
```

We can even use both:

```js
export const functionOne = () => {}
export const functionTwo = () => {}

export default functionOne
```

If you look at React itself it export as default the React and uses name export for another functions. eg.:

```
import React, { useState } from 'react';
```

src/Views/UsersPage/UserList/UserList.tsx
marcelo-aleixo: Nope, I would use another thing that can uniquely identify that component. If we use index, and let's imagine that we want to reorder this array of something, on reorder the component will receive another number as key therefore this component will re-render even thought the data itself wasn't modified.

src/Views/UsersPage/UserList/__tests__/UserList.test.tsx
marcelo-aleixo: I would say that is awesome, I really mean it! I am also aware that even though I think it is awesome I didn't do it in this project, but I left it as a TODO item that needs to be addressed.

src/Views/UsersPage/UserList/UserCard/UserCard.tsx
marcelo-aleixo: Right off the bat I would say that whenever we use React.FC it implies that the component can receive a prop children, even though we can or not use the prop inside the component itself.

src/Views/UsersPage/UserList/UserCard/UserCard.tsx
marcelo-aleixo: I would say yes, it is going to perform a shallow compare on the user prop, and detect if the user changed or not therefore not re-render if not necessary.

src/Views/UsersPage/UserList/UserCard/UserCard.scss
marcelo-aleixo: I would agree with you!

I even would create new variables like:

```scss
$smallFontSize: 12px;
$mediumFontSize: 14px;
$baseFontSize: 16px;
$largeFontSize: 20px;
$x-largeFontSize: 24px;
$xx-largeFontSize: 28px;
```

We can even go further and don't use fixed sizes like:

```css
margin-bottom: 16px;
```

instead add those variables:

```scss
$mobileBlockSize: 12px;
$desktopBlockSize: 16px;
```

and use multipliers like:

```scss
element: $desktopBlockSize;
element: $desktopBlockSize * 0.25;
element: $desktopBlockSize * 0.5;
element: $desktopBlockSize * 0.75;
...
element: $desktopBlockSize * 1.5;
element: $desktopBlockSize * 2;
...
```

This is the pattern I used in one of the projects I worked on, and I enjoyed doing so.

src/Views/UsersPage/UserList/UserCard/UserEdit/UserEdit.tsx
marcelo-aleixo: I like the `<Modal.Footer>` idea, this is something that I would change in this PR.
You reminded of a library called `AntD` that uses this pattern.

I would go on option `2.`

src/__snapshots__/storybook.test.ts.snap
marcelo-aleixo: It is nice to test presentational components, since we don't have any behavior to test. 
Since storybook has this plugin to automatically generate the snapshots tests, I thought that it was a good idea to add to the project.

src/components/modal/Modal.tsx
marcelo-aleixo: I would remove the scroll from the body, like I am doing in this `useEffect`:

```js
useEffect(() => {
    if (isOpen) {
      document.body.style.overflow = 'hidden';
    }
    return () => {
      document.body.style.overflow = 'scroll';
    };
  }, [isOpen]);
```

src/hooks/useScrollPosition.ts
marcelo-aleixo: We shouldn't use `sessionStorage` if we want to persist the data after the user closes the window/tab/browser.

For the scenario above, I judged the session storage as a good option since I want the user to be on top of the page when the re-opens the website.

src/GraphQL/mutations/UpdateUser.ts
marcelo-aleixo: I'm not sure if I fully understood what you're questioning, but I will try to answer, feel free to ask for more clarification.:

```js
  const update = useCallback((user: UpdateUserInput) => updateUser({
    variables: user,
  }), [updateUser]);
```

The updateUser is sending the whole user object as variable, even though the mutation will only use four of them.
This is not a problem since the mutation will not pass the other variables besides the four it is expecting.

marcelo-aleixo:
> Hi Marcelo! Thanks for the PR and the time you have spent on this.
> I left a few comments so we can have a little discussion, if you have any questions - feel free to ask 🙂

Hi @mat-sf Thank you for your CR, I answered all your comments, feel free to ask more! 😀

> Thanks for your submission! I left some questions, looking forward to your response :)

Hi @marcin-piela-sf Thank you for you CR, I answered your comments except two of them, I will put my head to rest a little bit, I will revisit your comments in the morning, feel free to add more comments! 😀

src/components/modal/Modal.tsx
marcelo-aleixo: Hey @marcin-piela-sf thank you for your patience, I've just updated this component to animate the modal on close, now I am keeping the overlay background for 3ms while I push the modal off the screen.

You will also see that I am not using the flag `isOpen` to create the portal when it is true, I'm now using a small library called `react-transition-group` that helps with css transitions.

I'm passing the flag `isOpen` as the value of the prop `in` of this library and also unmounting the component on exit, while keeping all animations.

src/App.test.tsx
marcelo-aleixo: I am not sure how to answer this question, but I would be happy to know some approaches to accomplish what you're asking for, would you mind helping me with this?

marcin-piela-sf: Yes sure, directly from documentation https://testing-library.com/docs/react-testing-library/setup/#custom-render :) 
Just import `render` from your project, with an additional setup instead of importing it from the library directly.

marcelo-aleixo: That is a really neat!! Thank you!

PR #43, opened 2021-09-30
Michael Cowan - FullStack Coding Challenge (Frontend Focused)

mikeysoftware:


package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "superformula-fullstack",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
```
mat-sf, line 5: I see some of the dependencies are in "devDependencies" and some in "dependencies" - why "@types/node" for example is in dependencies (I know some of those deps are here from CRA by default, just wanted to hear your thoughts)?

package.json
```diff
@@ -0,0 +1,78 @@
...
+    "@types/react-dom": "^17.0.0",
+    "date-fns": "^2.24.0",
+    "framer-motion": "^4.1.17",
+    "graphql": "^15.6.0",
+    "mapbox-gl": "^2.4.1",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-router-dom": "^5.3.0",
+    "react-scripts": "4.0.3",
+    "styled-components": "^5.3.1",
```
mat-sf, line 21: Can you elaborate on why did you choose styled-components? Do you know any other CSS-in-JS libs?

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default function Loader() {
+  return <LoaderWrapper />;
+}
+
+const LoaderWrapper = styled.div`
+  border: 0.375rem solid rgba(0, 0, 0, 0.125);
+  border-top: 0.375rem solid rgba(0, 0, 0, 0.5);
+  border-radius: 50%;
+  width: 2.75rem;
```
mat-sf, line 11: Can you elaborate more on why you use rem unit for width/height?

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,24 @@
...
+  height: 2.75rem;
+  margin-top: 0.375rem;
```
mat-sf, line 13: Can you elaborate more on why you use rem unit for margins?

src/components/UserCard/UserCard.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  p {
+    margin-top: 0.5rem;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  ${(props) =>
+    props.isHovered === true &&
```
mat-sf, line 121: Would it be possible to make it work only with CSS?

src/components/UserModals/CreateModal/CreateModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  return (
+    <Modal isModalVisible={isOpen} onModalClose={onClose}>
+      <CreateModalContainer>
+        <div className="header">
+          <h1>Create User</h1>
```
mat-sf, line 55: How in the production app do you usually write such strings in app? Directly in code?

src/components/UserModals/CreateModal/CreateModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+              id="description"
+              name="description"
+              type="text"
+              value={userData.description || ""}
+              onChange={handleInputChange}
+              placeholder="Brief Description"
+            />
+          </form>
+        </div>
+        <div className="footer">
```
mat-sf, line 92: I can see sometimes you use styled components, but sometimes just class names, can you plase elaborate more why?

src/views/UsersPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+    .loader {
+      display: flex;
+      justify-content: center;
+    }
+
+    .user__list {
+      display: grid;
+      justify-content: space-around;
+      grid-template-columns: repeat(auto-fill, minmax(25rem, 1fr));
```
mat-sf, line 128: I see sometimes you use grid, sometimes flex, do you have any general convention where you prefer one over the other?

src/views/UsersPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+import Loader from "../components/Loader";
+import useDebounceState from "../hooks/useDebounceState";
+import useSearchParams from "../hooks/useSearchParams";
+import Alert from "../components/Alert/Alert";
+import { NetworkStatus } from "@apollo/client";
+
+// Constants
+const PAGE_SIZE = 6;
+
+export default function UsersPage() {
```
mat-sf, line 19: WDYT about separating logic from dump/presentational components?

mat-sf (commented):
Hi Michael! Thanks for the PR and the time you have spent on this.
I left a few comments so we can have a little discussion, if you have any questions - feel free to ask 🙂

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,97 @@
+import ReactDOM from "react-dom";
+import { MouseEvent, useEffect } from "react";
+import styled from "styled-components";
+import { AnimatePresence, motion } from "framer-motion";
```
nickk-sf, line 4: A task spec asked not to use any external libs for animation but rather to implement one yourself. How would you do that?

src/components/UserModals/CreateModal/CreateModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { User } from "../../../graphql/schema.types";
+import { createUser } from "../../../graphql/mutations";
+
+// Components
+import Modal from "../../Modal";
+import Button from "../../Button";
+import Input, { Label } from "../../Input";
+import Alert from "../../Alert";
+
+// Interface
```
nickk-sf, line 13: Is there's anything valuable / important this comment tells us that we cannot figure out from the code itself?

src/components/UserModals/CreateModal/CreateModal.tsx
```diff
@@ -0,0 +1,144 @@
+import { useState, ChangeEvent, FormEvent } from "react";
+import styled from "styled-components";
+import { useMutation } from "@apollo/client";
+import { User } from "../../../graphql/schema.types";
+import { createUser } from "../../../graphql/mutations";
+
+// Components
```
nickk-sf, line 7: Is there's anything valuable / important this comment tells us that we cannot figure out from the code itself? What is your way of using comment (when and why)?

src/components/UserModals/EditModal/EditModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  return (
+    <Modal isModalVisible={isOpen} onModalClose={onClose}>
+      <EditModalContainer>
+        <div className="header">
+          <h1>Edit User</h1>
+        </div>
+        {/* Alerts */}
+        <div className="alerts">
+          {data?.updateUser && <Alert variant="success">{"SUCCESS :: User successfully updated."}</Alert>}
+          {error?.graphQLErrors && <Alert variant="error">{"ERROR :: Failed to update User."}</Alert>}
```
nickk-sf, line 90: How do you usually handle server errors? Any practices you follow?

src/hooks/useSearchParams.tsx
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from "react";
+import { useLocation } from "react-router-dom";
+// import { useHistory } from "react-router-dom";
+
+export default function useSearchParams(key: string, defaultValue: any) {
+  // State and Setter for debounced value
+  const location = useLocation();
+  // const history = useHistory();
```
nickk-sf, line 8: Do you usually keep commented code in a production app's repo?

src/views/UsersPage.tsx
```diff
@@ -0,0 +1,138 @@
...
+          {/* User Search */}
+          <div className="user__search">
+            <h1>Users List</h1>
+            <Input value={searchKeyword} onChange={(e) => setSearchKeyword(e.target.value)} placeholder="Search..." />
+          </div>
+          {/* Content */}
+          <div className="content">
+            {/* Alert */}
+            {!loading && data?.listUsers?.items.length === 0 && <Alert>{"INFO :: There are currently no users in the database."}</Alert>}
+            {!loading && error?.graphQLErrors && <Alert variant="error">{"INFO :: There are currently no users in the database."}</Alert>}
```
nickk-sf, line 74: How do you usually manage duplicated strings in a code base?

nickk-sf (commented):
I appreciate the time you put into this coding challenge @mikeysoftware 👏 Lots of work have been done. In addition to some inline comments I left, I've also got some high level observations regarding the working app that I'd like to know more about:

1. When loading the app I can see two separate requests for `GetLocationInformation`. Is that intentional? Why do we need to calls even if we didn't open Edit Modal yet which uses that address?

2. How'd you implement this part of requirements?

    > pagination state should be present in URL query (eg ?page=1) to allow for behavior where the user can reload the page while still returning to their current scroll position

3. There's no way to close the modal after successfully saving details. And there's a success alert showing up when opening the profile to edit again. Is that how you usually handle modal interactions? What would you change?
    image

4. When filtering the list and clicking "Load More", I can see some items get rendered but then disappear. This may confuse your end users. How would you improve that? What technique you could utilize to get rid of such client-side glitches? 

    One extra observation: Each time we click "Load more", there's 2 "GetLocationInformation" triggering along with an expected "ListUsers" request. Why do we need those duplicated requests for location info? Is this even used for when we just want to list users?

package.json
mikeysoftware: As you speculated, some dependencies are in "dependencies" due to using the CRA —typescript template. Type definitions I installed myself are in "devDependencies". While modern bundlers only compile what is used (tree-shaking), organizing the dependencies in their correct groups allows for other users as well as other CLI's commands and flags to correctly identify what is required to run in production, used for development, and/or is peer dependent.

package.json
mikeysoftware: Styled-Components was a decision made out of preference since I was allowed the freedom. My path in styling has been CSS → Sass → Styled Components, for the most part. Other than using StyleSheets in react-native intermittently, my CSS-in-JS library of choice is SC due to its features, popularity, and great community support. I am completely open-minded when it comes to tooling though.

src/components/Loader/Loader.tsx
mikeysoftware: Originally, my response to this question would be similar to the one above, however, you asking it twice raised my curiosity. After some research, I found this [article](https://medium.com/@sascha.wolff/dont-use-rem-em-for-paddings-margins-and-more-94e19026b000) that made some interesting points about using "rem" for most properties. . The writer argues that using rem units for margins and paddings creates friction with the browsers accessibility settings; mainly the "Font Size" setting in preferences. I'm not completely sure if this is the direction you were pushing me in, but I'd love to hear your thoughts as well.

src/components/UserCard/UserCard.tsx
mikeysoftware: Yes, it is. By using the parent element and pseudo-classes, I could have done something like the following:
```css
.container:hover .actions .container:hover .about .created { 
    opacity: 1;
}
```
This uses CSS specificity to target the required classes. I think at the time I went this route because I had not decided on whether to use conditional rendering or opacity/visibility.

src/components/UserModals/CreateModal/CreateModal.tsx
mikeysoftware: Yes, provided that the component can't be narrowed down further into a template structure. At that point a string or render prop could be specified. Unless you are referring to internationalization, in that case; dynamic variables or the use of a trusted i18n library are the most suitable choices in my opinion.

src/components/UserModals/CreateModal/CreateModal.tsx
mikeysoftware: This was done mostly for the speed of development at the time. I was quickly scaffolding the UI components to test design accuracy. A mixture of both methods is an anti-pattern and bad practice. In review/refactoring, I would like to hope this code quality issue would be caught and rectified.

src/views/UsersPage.tsx
mikeysoftware: As opposed to a general convention, I assess each layout individually on a case-by-case basis to pick the better option. In this situation, the list container seemed to be more focused on content layout and had a clearly defined pattern in columns and rows. I chose CSS Grid for its simplicity when working with grids and grid gaps. While I am aware that the same can be achieved with Flexbox and some intuitiveness regarding margins, I felt like CSS Grid was a more clear-cut choice on this occasion.

src/views/UsersPage.tsx
mikeysoftware: It is always recommended to de-couple and separate the concerns of logic in code. This is an oversight that I wish to thank you for catching, as It would be more ideal to pass the preferred "PAGE_SIZE" as a property to the component, or make it configurable via a variable.

.env
```diff
@@ -0,0 +1,6 @@
+# AWS - AppSync + Lambda + DynamoDB
+REACT_APP_APPSYNC_API_URL=https://fdldn2jrvvdippv2lczxpf3fcu.appsync-api.us-east-2.amazonaws.com/graphql
+REACT_APP_APPSYNC_API_KEY=da2-bnaosfk73jaqvidiq7fpbtgh3y
+
+# MAPBOX
+REACT_APP_MAPBOX_ACCESS_TOKEN=pk.eyJ1IjoibWlrZXlzb2Z0d2FyZSIsImEiOiJja3R6NnNyaTQwaHZjMnVwaTQ1aWUyeGM2In0.6XX8gkPPnW0LUt-IOO0zlQ
```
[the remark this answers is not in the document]
mikeysoftware: Not at all. I am fully aware of the security risks involved with handling access tokens and secrets in a production application and take the correct measures necessary to combat such a breach. However, for ease of demonstration and startup, I had mentioned in the README.md under Available Scripts, the following: 
```
NOTE: ".env" file was committed for ease of start-up.
```
In contrast, there is a ".env.example"  that more accurately depicts the type of configuration template file that is usually committed. These tokens are both public and limited in use strictly for this coding challenge.  I appreciate your concern.

cypress/integration/user-card/edit-details.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+    });
+
+    it('can edit user details', () => {
+        cy.wait(2000);
+        cy.get("#edit-modal-trigger").first().click();
+
+        cy.get("#name").clear().type("Cypress Test User");
+
+        cy.get("#edit-form").submit();
+        cy.wait(1000);
```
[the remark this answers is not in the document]
mikeysoftware: A more reliable and recommended way for handling requests is to use one of the many testing strategies cypress provides for such a use case. You can rely on the real request responses/timing that is more indicative of real-world results by asserting on-network calls or by stubbing the requests for more testing control using the "intercept()" method.

package.json
```diff
@@ -0,0 +1,78 @@
...
+    ]
+  },
+  "browserslist": {
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
```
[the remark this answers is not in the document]
mikeysoftware: I support all modern browsers on production projects; including Chrome, Edge, Firefox, and Safari.
In addition to manual testing and accessibility tools, I reference public projects like CanIUse and CSS Reset to validate feature availability and ensure consistency.

src/components/Alert/Alert.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import { Story } from "@storybook/react";
```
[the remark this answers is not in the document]
mikeysoftware: I have used Storybook before. The main advantages include developing a system or guide with live component documentation and easy-to-use visual testing and component interaction. While there are other benefits, these stand out in a team and production environment where a source of truth and guidance is needed to synchronize teams.

src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { render, screen } from "@testing-library/react";
+
+import "@testing-library/jest-dom/extend-expect";
+
+import { Info, Success, Error } from "./Alert.stories";
+
+it("renders alert with 'info' styles", () => {
+  render(<Info {...Info.args}>{Info.args?.children}</Info>);
+  expect(screen.getByRole("alert")).toHaveStyle("background-color: #d4d4d4");
```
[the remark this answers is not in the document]
mikeysoftware: The color was used in this simple component as there was no other real indicator of status. However, it could be refactored to use a custom attribute or follow a standard in error state text that a regular expression could be used to identify.

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const AlertContainer = styled.div<AlertProps>`
+  padding: 2rem 4rem;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  font-size: 1.5rem;
+  line-height: 2rem;
```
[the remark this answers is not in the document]
mikeysoftware: Actually, no. It is recommended to use a unitless number when dealing with line-heights as the browser preserves the font-size/line-height ratio even if the "font-size" changes. Therefore a recalculation or tweaking would not be necessary to keep the same level of legibility.

src/components/Alert/Alert.types.ts
```diff
@@ -0,0 +1,6 @@
+interface AlertProps {
```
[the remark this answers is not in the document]
mikeysoftware: Type aliases and interfaces are very similar and in most cases can be alternated. The key differentiation noted by the official TypeScript documentation is that a type can't be re-opened/redefined to add new properties while an interface is always extendable. Therefore I have built the habit of using interfaces in most situations just in case that situation may arise.
I am continuously making discoveries in TypeScript so please feel free to share your thoughts and recommendations regarding the matter.

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,16 @@
+import styled from "styled-components";
+import AvatarProps from "./Avatar.types";
```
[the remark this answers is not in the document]
mikeysoftware: It is recommended to declare and name exports. I did not think it was necessary in this case but it is not a good practice. Default exports assume there is only one export from the module and therefore you might import the wrong value. Properly naming and exporting types in a declarative way prevents bugs, improves readability and overall code quality.

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function Button({ variant = "primary", loading, children, ...rest }: ButtonProps) {
+  return (
+    <ButtonWrapper variant={variant} {...rest}>
+      <ButtonContainer>{loading ? <Loader /> : children}</ButtonContainer>
+    </ButtonWrapper>
+  );
+}
+
+// Styled
```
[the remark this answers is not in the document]
mikeysoftware: It really depends on the size or complexity of the project and I always try to stick to one or the other. Later in the lifecycle of the application, I started extracting elements of the component into their respective files in the same folder. ( see .stories, .test, .types files). This could be further extended to have a ".styled" file to separate the files, similar to how CSS modules are handled in projects to avoid namespace collision and organize the project.

src/components/Card/Card.types.ts
```diff
@@ -0,0 +1,7 @@
+import { HTMLMotionProps } from "framer-motion";
+
+interface CardProps extends HTMLMotionProps<"div"> {
+    children: any;
```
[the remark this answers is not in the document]
mikeysoftware: A "JSX.Element" or "React.ReactNode" is recommended for handling the "children" property.  The key difference being that "React.ReactNode" allows for strings and other children that are typically provided as children, not just elements. But as per type safety, this is solely based on the use-case of the component and what should be "acceptable" as a property. This use of "any" was not intended to make it to production.

src/components/Map/Map.tsx
```diff
@@ -0,0 +1,57 @@
...
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import mapboxgl from "!mapbox-gl";
+import MapProps from "./Map.types";
+import "mapbox-gl/dist/mapbox-gl.css";
+
+const { REACT_APP_MAPBOX_ACCESS_TOKEN } = process.env;
+mapboxgl.accessToken = REACT_APP_MAPBOX_ACCESS_TOKEN || "";
+
+export default function Map({ mapCenter = [0.0, 0.0] }: MapProps) {
+  const mapContainer: any = useRef(null);
```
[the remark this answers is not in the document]
mikeysoftware: This is another oversight of the usage of "any" when debugging/developing component logic. I hope I have demonstrated the correct use of type definitions elsewhere in the project.

src/components/Map/Map.tsx
```diff
@@ -0,0 +1,57 @@
...
+      });
+      new mapboxgl.Marker({ color: "#dc2626" }).setLngLat(mapCenter).addTo(map.current);
+    }, 350);
+  });
+
+  // Update Map Center + Marker
+  useEffect(() => {
+    if (!map.current) return;
+    map.current.setCenter(mapCenter);
+    new mapboxgl.Marker({ color: "#dc2626" }).setLngLat(mapCenter).addTo(map.current);
```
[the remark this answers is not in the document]
mikeysoftware: In the same way that a check was made to ensure multiple instances of the map did not exist, a check should be added for the markers as well. It would be more efficient to simply re-use the existing marker by accessing the Mapbox GL API and changing its coordinates. This improvement can be made.

src/components/Modal/Modal.tsx
mikeysoftware: Framer motion provides a simple and intuitive API for handling on-mount and unmounting animations, however, there are ways to provide similar functionality without external dependencies.

In terms of logic, a High-Order-Component could be used to abstract the logic responsible for controlling a delay in mounting and un-mounting animations. (A combination of "useEffects", "setTimeouts", and CSS Animations or the Web Animation API could be used to achieve this.) I always follow the [FLIP technique](https://css-tricks.com/animating-layouts-with-the-flip-technique/) when implementing animations for web interfaces.

src/components/UserModals/CreateModal/CreateModal.tsx
mikeysoftware: No, the existing code is clear and declarative. These comments were purely for development purposes and are to be removed.

src/components/UserModals/CreateModal/CreateModal.tsx
mikeysoftware: Please see the above comment. This is a similar situation.

src/components/UserModals/EditModal/EditModal.tsx
mikeysoftware: Apart from catching the errors as I tried to do in the query calling function (using try-catch), there are improvements that can be made for handling server errors. 
a) Using an Error Boundary to capture errors before accidentally disrupting the user experience.
b) Utilizing retries to attempt re-performing the intended functionality. (Up to a limit)

src/hooks/useSearchParams.tsx
mikeysoftware: No, any comments unless specifically mentioning an important note are to be removed either manually or using a git hook to detect and identify them.

src/views/UsersPage.tsx
mikeysoftware: While there is currently no fallback in this coding challenge, this type of error is usually handled by multiple layers of the development process. Unit tests, E2E tests, Code Review, QA, and Regression Testing are usually thorough enough to prevent any issues arising from this.

mikeysoftware:
> I appreciate the time you put into this coding challenge @mikeysoftware 👏 Lots of work have been done. In addition to some inline comments I left, I've also got some high level observations regarding the working app that I'd like to know more about:
> 
> 1. When loading the app I can see two separate requests for `GetLocationInformation`. Is that intentional? Why do we need to calls even if we didn't open Edit Modal yet which uses that address?
> 2. How'd you implement this part of requirements?
>    > pagination state should be present in URL query (eg ?page=1) to allow for behavior where the user can reload the page while still returning to their current scroll position
> 3. There's no way to close the modal after successfully saving details. And there's a success alert showing up when opening the profile to edit again. Is that how you usually handle modal interactions? What would you change?
>    image
> 4. When filtering the list and clicking "Load More", I can see some items get rendered but then disappear. This may confuse your end users. How would you improve that? What technique you could utilize to get rid of such client-side glitches?
>    One extra observation: Each time we click "Load more", there's 2 "GetLocationInformation" triggering along with an expected "ListUsers" request. Why do we need those duplicated requests for location info? Is this even used for when we just want to list users?
>    

Great questions, I will try to answer them to the best of my current knowledge.

1. This is to be expected but not intentional. This is a side effect of the subpar design choice you have pointed out. Currently, the modals could be improved in multiple ways, but regarding your point, here are two solutions I see that could remedy the issue of premature calls:
a) Move the logic to the parent and pass it to the modal as an event function that can be called on-mount or wherever else required.
b) Use conditional rendering on the modal itself instead of its children so that the logic isn't processed until the component is visible/active.

2. The URL query functionality was implemented by taking advantage of the URLSearchParams API that is available natively.  I had originally used it directly but later then added React-Router to the project when creating the main page/view. In its latest revision, I utilized the "useLocation" hook to access the URL's parameters and implement the page functionality accordingly. Please try adding "?page=2" on the URL's path to test the logic. But now after trying it myself, it seems I may have forgotten to revisit the requirement as I remember facing challenges when trying to update the existing query string programmatically. While the logic exists, the URL path does not accurately represent the state in some cases. I will remedy this user-experience problem at the earliest opportunity.

3. The success state would be improved/corrected along with the refactoring suggestions I made above. (see response to first question). Regarding exiting the modal, however, there are three options implemented to terminate the dialog.
a) Clicking the cancel button will close the modal.
b) Clicking the overlay/background will close the modal. (Opinionated choice)
c) Pressing the "exit" key while the modal is in focus/active.

4. I was unaware of this bug as I believe it was functional during testing. However, there are points to be investigated. The "Load More" event handler takes advantage of the "fetchMore()" method provided by the "useQuery" hook from Apollo Client. Maybe a race condition is happening where I provided the merge function for caching existing results in the Apollo Provider. Further debugging will be necessary to report a succinct solution.

PR #45, opened 2021-11-14
Pull Request for UI Context Exam

hudson21:
Hello Superformula Team, my name is Carlos Hudson and I am a candidate who is applying to a Senior Frontend position you have. In this PR you can find all the requirements listed in the UI Context exam. There is a thing I tried doing but sometimes is not working that it is to save the scroll position once new data users are fetched from the backend, but apart from that everything else I think is handled. 
I did the application using Next.js and created the backend server with apollo-server-micro library that is focused on Next.js ecosystem. I used MongoDB instead of DynamoDB because it was the DB I have more experience with so far but I always like to learn new things that gets me excited. 
I focused more on the UI requirements than backend despite of creating the backend from scratch. 
If there is something you would like me to change or any suggestion, feel free to ping me and I am always happy to get feedback from experienced people to always continue learning :)


ui-context-exam/components/UI/Avatar/index.js
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <img
+      className={classes.avatar}
+      alt={`${name} image`}
+      src={BASE_UNSPLASH_URL(id)}
+    />
+  );
+};
+
+Avatar.propTypes = {
```
nickk-sf, line 15: Have you used TypeScript and how you would improve typings over propTypes?

ui-context-exam/components/UI/Avatar/styled.module.css
```diff
@@ -0,0 +1,7 @@
+.avatar {
```
nickk-sf, line 1: What other CSS methodologies have you used before? Which is your go-to tool?

ui-context-exam/components/UI/Button/Button.test.js
```diff
@@ -0,0 +1,40 @@
...
+      value: 'Load More',
+      isPrimary: true,
+      onClick: jest.fn(),
+    };
+
+    render(<Button {...expectedProps} />);
+  });
+
+  it('Should display value passed as props', () => {
+    const button = document.querySelector(
```
nickk-sf, line 23: What are the preferred ways to query for elements and what would you improve in your test knowing that?

ui-context-exam/components/UI/Button/styled.module.css
```diff
@@ -0,0 +1,58 @@
+.button {
+  text-transform: uppercase;
+  width: 280px;
+  height: 90px;
+  padding: 11px 14px;
+  font-weight: 600;
+  font-size: 24px;
```
nickk-sf, line 7: What are your thoughts on using `px` for `font-size`? Is it how you write styles?

ui-context-exam/components/UI/Card/styled.module.css
```diff
@@ -0,0 +1,93 @@
...
+
+.card img {
+  width: 168px;
+  height: 168px;
+  border-radius: 50%;
+  margin-bottom: 32px;
+  border: 1px solid #979797;
+}
+
+.card h2 {
```
nickk-sf, line 43: if the markup changes, this means you'll need to remember to adjust your styles as well. How do you deal with that knowing you have lots of styles based on element selectors?

ui-context-exam/components/UI/CardList/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+const CardList = ({ users, openUpdateModal, deleteUserHandler }) => {
+  return (
+    <>
+      {users.length > 0 ? (
+        <div className={classes['card-list']}>
+          {users.map((user) => (
+            <Card
+              key={user._id}
+              openUpdateModal={() => openUpdateModal(user)}
```
nickk-sf, line 14: generally, is it a good idea to pass reference values as React component props? What are the potential drawbacks?

ui-context-exam/components/UsersPage/index.js
```diff
@@ -0,0 +1,217 @@
...
+  useEffect(() => {
+    return () => {
+      handleScrollPosition();
+    };
+  }, []);
+
+  const handleScrollPosition = () => {
+    const scrollPosition = sessionStorage.getItem('scrollPosition');
+    if (scrollPosition) {
+      window.scrollTo(0, +scrollPosition);
```
nickk-sf, line 73: What do you think of using magic tricks like `+` on production code? Any ways to express code better?

ui-context-exam/graphql/Mutation.js
```diff
@@ -0,0 +1,83 @@
...
+      updatedAt: Date.now(),
+    };
+
+    try {
+      const createdUser = await insertDocument(client, 'users', userToCreate);
+      userToCreate._id = createdUser.insertedId;
+
+      return userToCreate;
+    } catch (e) {
+      throw new Error('User could not be created.');
```
nickk-sf, line 23: Will this be helpful in production? What else do you usually do when handling server errors?

ui-context-exam/graphql/gql/mutations/CREATE_USER.js
```diff
@@ -0,0 +1,15 @@
+import { gql } from '@apollo/client';
+
+export const CREATE_USER = gql`
+  mutation CreateUser($data: CreateUserInput!) {
```
nickk-sf, line 4: Have you used GraphQL before? What is your experience?

ui-context-exam/next.config.js
```diff
@@ -0,0 +1,45 @@
...
+      return [
+        {
+          source: '/',
+          destination: '/1',
+        },
+      ];
+    },
+    env: {
+      MONGODB_USERNAME: 'hudson21',
+      MONGODB_PASSWORD: 'xD6yVkQejkE3kTaG',
```
nickk-sf, line 38: How do you ensure your secrets aren't leaked in prod apps?

nickk-sf (commented):
Hey @hudson21 First of all, thanks for your time and all the effort you put into the test assignment. We checked each and every feature you developed. And special thanks for a detailed README, it helped a lot on navigating around. 👍 

I left some inline comments to learn more about your experience. Besides, here're some extra questions I have for you.

1. Sometimes, I noticed that when performing a search being on page > 1 and then removing search, it reveals `0` results dues to the wrong API request:

image

2. You seem to trigger api requests on each char typed into the search form input which doesn't seem like a really performant way of handling this. Can you think of anything to improve performance here?

3. How have you used Next.js before? Have you worked (wrote) ssr apps?
4. What experience do you have with AWS? Which services have you used previously?

Again, thanks for your time and looking forward to speaking to you!

hudson21:
Hello @nickk-sf, thank you very much for your feedback. I will give you an answer for every question you asked me:

1. Sometimes, I noticed that when performing a search being on page > 1 and then removing search, it reveals 0 results dues to the wrong API request
R/ You are totally right. The problem with this was that I set a redirect configuration on the `next.config.js` file but it seems this configuration was not taken on production environment. The normal behavior should be that as soon as you hit the home page `https://carlos-hudson-superformula-exam.vercel.app/` or `http://localhost:3000` you should be redirected to the `BASE_URL/1`

2. You seem to trigger api requests on each char typed into the search form input which doesn't seem like a really performant way of handling this. Can you think of anything to improve performance here?
R/ I could have placed a submit button to trigger the search filter instead of filtering on every chart typed. On this way I could have avoided refetching every time there is a change.

3. How have you used Next.js before? Have you worked (wrote) ssr apps?
R/ Yes I am aware of the advantages that Next.js offers like Static Generation, Server-Side Rendering and Client Side Rendering as well coming from React. I could have used getServerSideProps in order to get the latest data on every new request made by the client 

4. What experience do you have with AWS? Which services have you used previously?
R/ I have used AWS a little bit on services like S3, EC2 and a Lambdas. But since I did not have that much expertise on the services asked in the backend side that is why I decided to stick with the technologies I knew so far

5. Have you made web apps accessible before? What is your typical approach in handling that?
R/ I am pretty aware of accessibility web apps and how they impact the way screen readers can read your page. Some of the approaches that you can take from the start of a project are:
- Semantic HTML: Which means use the specific tags needed for the elements you want to express on the DOM instead of just using pure divs to accommodate your layout
- Use Elements that have keyboards access like: `<input>, <select>, <textarea>, <button>, and <object>` are focusable if they are not disabled.

ui-context-exam/components/UI/Avatar/index.js
hudson21: Yeah I have previously used TypeScript on the Angular ecosystem and I know I could have used interfaces over propTypes to validate my object structure ahead of time. I decided to cope with PropTypes over TypeScript interfaces because they helped me for a faster setup of the storybook components, but I know this was just a decision based on my personal perspective. This could have changed if this application was on an enterprise environment.

ui-context-exam/components/UI/Avatar/styled.module.css
hudson21: I have used `BEM` (Block, Element, Modifier) previously and I am aware there are others like `Object-Oriented CSS (OOCSS), Scalable and Modular Architecture for CSS (SMACSS)`  but I have not used these last ones before.

PR #50, opened 2021-12-13
Test task

ivan-shaban:
Hello, detail description you may find in `README.md` file.


src/components/pages/users/UsersPage.tsx
```diff
@@ -0,0 +1,143 @@
...
+        data, fetchMore
+    } = useQuery<{ users: User[] }, FindManyUserArgs>(GET_ALL_USERS_QUERY, {
+        errorPolicy: 'all',
+        notifyOnNetworkStatusChange: true,
+        variables: {
+            // @ts-ignore
+            take: pageIndex * process.env.NEXT_PUBLIC_USERS_PER_PAGE,
+        },
+    })
+    const filteredData = useMemo(() => {
```
surbina-sf, line 45: Is there any other way to implement search/filtering? If so, what are the benefits/downsides compared to the approach you have taken?

src/pages/api/graphql.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const cors = Cors()
+let apolloServer: ApolloServer
+
+const getServer = async () => {
+    if (!apolloServer) {
+        const context = (): Context => ({prisma})
+        const schema = await buildSchema({
+            resolvers: crudResolvers,
+            emitSchemaFile: path.resolve(__dirname, '..', '..', '..', '..', 'generated-schema.graphql'),
```
surbina-sf, line 23: Are there any drawbacks to using the autogenerated schema like this?

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,33 @@
...
+    Primary = 'primary',
+    Danger = 'danger',
+}
+
+export interface Props extends Omit<DetailedHTMLProps<ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>, 'type'> {
+    readonly isDisabled?: boolean
+    readonly type?: ButtonType
+}
+
+export const Button: FC<Props> = memo(({
```
surbina-sf, line 20: How does wrapping a component in `React.memo` impact the performance of a component?

src/components/Button/__tests__/Button.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const defaultProps: Props = {}
+
+const createTree = (props?: Partial<Props>) => {
+    return renderer.create(<Button {...defaultProps} {...props}/>).toJSON()
+}
+
+it('default', () => {
+    const tree = createTree()
+    expect(tree).toMatchSnapshot()
```
surbina-sf, line 17: What are the downsides of snapshot testing?

src/components/pages/users/components/EditUserInfoOverlay/EditUserInfoOverlay.module.scss
```diff
@@ -0,0 +1,224 @@
+@import "../../../../../styles/all";
+
+$default_margin: 16px;
+
+.layout {
+    display: block !important;
+    background-color: rgba(0, 0, 0, 0);
```
surbina-sf, line 7: How could you avoid repeating these constant values all the time and also be sure the same colors are used across the application?

package.json
```diff
@@ -0,0 +1,88 @@
...
+    "prisma:generate-client": "npx prisma generate",
+    "prisma:studio": "npx prisma studio",
+    "storybook:start": "start-storybook -p 6006",
+    "storybook:build": "build-storybook",
+    "postinstall": "npm run prisma:generate-client",
+    "predev": "npm run prisma:generate-client",
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "test": "jest --watch",
```
surbina-sf, line 14: How could you ensure tests are not accidentally broken when introducing some changes?

src/components/pages/users/components/EditUserInfoOverlay/EditUserInfoOverlay.module.scss
```diff
@@ -0,0 +1,224 @@
...
+        width: $card_min_width * 4 + $card_min_gap * 3;
+    }
+    @include four_cards_small_max_gap {
+        width: $card_min_width * 4 + $card_max_gap * 3;
+        padding: $default_margin * 2;
+    }
+    @include four_cards_big_max_gap {
+        width: $card_max_width * 4 + $card_max_gap * 3;
+        padding: $default_margin * 4;
+    }
```
surbina-sf, line 97: Could this be solved with flexbox/grid?

ivan-shaban:
I apologize, I have unforeseen problems with the sewerage system in the apartment, I will answer your questions during the weekend

src/components/pages/users/UsersPage.tsx
ivan-shaban: Sure, we can make direct server requests with filtering params and get all data in responses.

In that case we get next cons:

- ui response will be delayed because we should await response from server
- probably we should throttle requests, to increase server performance and reduce loading, also in most cases user don't want to search for every letter that he input, but only after he input some phrase
- need more logic to handle request\response ui states
- more complex logic for pagination, it's harder to use autogenerated solutions, because it looks like we need to implement custom resolver

And next pros:

- we may user more complex filtering logic at server side
- ui logic can be simplified

Examples:
```
...
    const {
        data, fetchMore, refetch,
    } = useQuery<{ users: User[] }, FindManyUserArgs>(GET_ALL_USERS_QUERY, {
        errorPolicy: 'all',
        notifyOnNetworkStatusChange: true,
        variables: {
            // @ts-ignore
            take: pageIndex * process.env.NEXT_PUBLIC_USERS_PER_PAGE,
            where: filterValue ? {
                OR: [
                    {
                        address: {
                            contains: filterValue,
                        },
                    },
                    {
                        name: {
                            contains: filterValue,
                        },
                    },
                    {
                        description: {
                            contains: filterValue,
                        },
                    },
                ],
            }: undefined,
        },
    })
...
    useEffect(() => {
        refetch()
    }, [filterValue, refetch])
```

src/pages/api/graphql.ts
ivan-shaban: We are restricted with generator engine features, it is hard to extend generated schemas \ code with some custom logic (but it really depends). If you need some unique customization, it is harder to deliver it to production, because you should add feature to engine and await while it would be published.

src/components/Button/Button.tsx
ivan-shaban: `React.memo` it is analog of `React.PureComponent`, it MAY increase ui performance, but official documentation also mention that `Do not rely on it to “prevent” a render, as this can lead to bugs.`. 

How it works: if you component get new props with old values to `React.memo` prevent new rendering and return last memoized value.

But it also has some drawbacks:

- it creates memoization function which consume some memory and other computer resources
- due to it is `hoc` it increase element tree nesting

src/components/Button/__tests__/Button.tsx
ivan-shaban: Here short list of cons:

- covers only component structure, it completely ignore css and inner logic which outside of test scope
- you should manually manage what should be mocked, to don't leak nested component's logic\structure to component snapshot, it easy to forget about it and you will get mess in snapshot result
- requires more time to complete then regular unit tests
- requires explicit dates mocking

Pros of that kind of testing:

- if configurated correct - very readable
- easy to maintain

src/components/pages/users/components/EditUserInfoOverlay/EditUserInfoOverlay.module.scss
ivan-shaban: We can extract some common values to variables and use them across whole application.

We can use `theme` ui pattern, which is all about using different style variables.

package.json
ivan-shaban: Sorry, but i don't understand your question :( If we talking about `how to prevent to deliver logic that breaks our tests into repository?`. In that case we may use some client side hooks, that run tests locally before commit\push and aborts operation if your local tests has failed. Also we may run the same validation at ci server when you create \ update your pull request.

src/components/pages/users/components/EditUserInfoOverlay/EditUserInfoOverlay.module.scss
ivan-shaban: Probably i don't understand your question again. We definitely can describe that layout with grid, but if we talking about flexbox so we need some nested containers which can has own layout logic.

Can you clarify your question?

ivan-shaban:
I answers all comments, but some of them need more clarifications.

surbina-sf:
@ivan-shaban  Thanks for your answers and the time you put into this challenge. We'll take a look at them and get back to you.

PR #53, opened 2021-12-23
Ahmed Tohamy - Frontend Context

AhmedTohamy01:
**_IMPORTANT NOTES (Please Read First)_**

1) While this PR targets _**ONLY**_ Frontend Context, and I'm _**NOT**_ interested to work as a Backend or FullStack Dev, But i believe that the frontend developer who knows what is happening in Backend is better than the frontend developer that doesn't know.  That's why i built a backend GraphQL server with the technology stack i'm familiar with (Node.JS, Express, MongoDB) to show you that i know how to build a basic backend server and create a GraphQL Schema.

2) I created this web application as a _**PIXEL PERFECT**_ web application and i used the Figma Design **_AS IS_**  (main page using a screen width of 1800px), so you may need to zoom-out your browser in order to view it exactly as the design if you don't have a big screen (I had to use 75% of my browser size to view it correctly on my HP laptop).

3) I didn't use Unspalsh API and i used another public images API which provides also random images, and the reason for this is that Unspalsh API now has a limit of 50 call/hour. And also Unspalsh API give us repeated images (The goal is just to have a random avatar, am i right?).

4) There is nothing in the Figma Design about how to create cards, I added a Big PLUS icon beside the Main title to be used for adding new cards.. 

5) The ReadMe file will contain the "User Guide" (How to use the app as a user, listing all app features in detail) and also will contain "Use in Local Development Guide" (Detailed steps to get the app up and running in your local development environment). 

**_See ReadMe file here_** >> https://github.com/AhmedTohamy01/test-web-stack/tree/ahmed-tohamy-frontend-context

Everything else you requested will be in this PR Description.

6) I added a Live Demo for this app >> https://superformula-react-graphql.vercel.app/

**_Check List, What has been done ?_**

- Build GraphQL API that can create/read/update/delete user data from a database. ................ **DONE**
- The search functionality should perform real time filtering on client side data and API side data ................ **DONE**
- List of users should be updated automatically after single user is updated ................ **DONE**
- Create modal from scratch - please don't use any library for it ................ **DONE**
- Modal should be animated ................ **DONE**
- Allow for behaviour where the user can reload the page while still returning to their current scroll position ................ **DONE**
- Typescript preferred ................ **DONE**
- End-to-End cypress tests ................ **DONE**
- Do not use any additional libraries with predefined styles like react-bootstrap, material-ui ................ **DONE**
- Write clear documentation on how the app was designed and how to run the code ................ **DONE**
- Providing an online demo is welcomed ................ **DONE**
- Provide a description of how you approach mobile friendly apps ................ **DONE (details below)**
- Write concise and clear commit messages ................ **DONE**
- Describe optimization opportunities when you conclude ................ **DONE (details below)**
- Handle server errors ................ **DONE**
- Handle loading states ................ **DONE**
- Delete user functionality ................ **DONE**

**_How you approach mobile friendly apps ?_**

Usually, I build the app in components, so when I build each component i build it completely from desktop view to small mobiles view, then move to build the next component. 

For example: building the Navbar first, making sure that it looks good in all targeted screen sizes (according to the design), and then starting building the sidebar with the same approach.

I do NOT build the whole app in the Desktop view first and then start customizing it. i do NOT use this approach. 

**_Optimization opportunities_** 

If i have time i will do the following:

- Make this web app fully responsive across all devices from big screens to small mobile devices. 
- Replace the static map image with real mapping API, so when the user writes an address in the address field it reflects in the map.
- Connect the address field to Google Places with auto complete.
- Add confirmation messages when the user create/update/delete card.
- Add validation messages in the Modals, so it tell the user that he can NOT leave name field without value for example.

**_Bonus Point: React App Map (To better understand how the react app is structured)_**

Superformula Map - Copy


cypress.json
```diff
@@ -0,0 +1,7 @@
+{
+  "ignoreTestFiles": ["**/2-advanced-examples/*", "**/1-getting-started/*"],
```
marcin-piela-sf, line 2: Why did you push examples of E2E tests into this PR? What was the reason? Just by mistake? :)

src/GraphQLQueries/GraphQLQueries.tsx
```diff
@@ -0,0 +1,66 @@
+import { gql } from '@apollo/client'
+
+export const GET_ALL_USERS = gql`
+  query ($filter: String!, $limit: ID!) {
+    users(filter: $filter, limit: $limit) {
```
marcin-piela-sf, line 5: Let's say that we want to filter by description and address as well. How would you change the current schema to provide this new functionality?

src/Pages/MainPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+  const [showEditModal, setShowEditModal] = useState(false)
+  const [showDeleteModal, setShowDeleteModal] = useState(false)
+  const [activeCard, setActiveCard] = useState<CardType | null>(null)
+  const [searchTerm, setSearchTerm] = useState('')
+  const [filterTerm, setFilterTerm] = useState('')
+  const search = useLocation()
+  const urlNum = search.pathname.split('=')[1]
+  const [limit, setLimit] = useState(parseInt(urlNum))
+  const history = useHistory()
+  const { loading, error, data } = useQuery(GET_ALL_USERS, {
```
marcin-piela-sf, line 31: With this approach, the `data` variable is not typed, so in the code, we don't know what is it (it's "any", am I right?). How would you fix it? How do you handle it in real projects?

src/Pages/MainPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+    const timeoutId = setTimeout(() => {
+      const handleFilterTerm = () => {
+        setFilterTerm(searchTerm)
+      }
+      handleFilterTerm()
+    }, 1000)
+    return () => clearTimeout(timeoutId)
+  }, [searchTerm])
+
+  useEffect(() => {
```
marcin-piela-sf, line 61: What is doing this `effect`? Could you elaborate more (from UX perspective)?

src/Pages/MainPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+    setTimeout(() => {
+      document.getElementById('search-box')?.focus()
+    }, 400)
+  }, [filterTerm])
+
+  if (loading)
```
marcin-piela-sf, line 67: In terms of UX, is it the best way of handling loading state?

src/Pages/MainPage.tsx
```diff
@@ -0,0 +1,245 @@
...
+    return (
+      <SpinnerWrapper>
+        <SorryImage src='/sorry.png' alt='shy girl' />I don't know how to say
+        this to you ... <br />
+        But, we couldn't contact the server to get data, please try again later
+        !
+      </SpinnerWrapper>
+    )
+
+  if (data.users.length === 0)
```
marcin-piela-sf, line 84: A difference between `users.length === 0` and the state when users exist is not so big, wouldn't be better to add one condition inside `CardWrapper` (line 141)?

src/Components/Card/Card.interfaces.ts
```diff
@@ -0,0 +1,8 @@
+import { CardType } from '../../utils/Shared.interfaces'
+
+export interface PropsType {
+  name: string
+  description: string
+  handleEditIconClick: (arg0: CardType) => void
```
marcin-piela-sf, line 6: Looks like `arg0` is not the best way for naming a variable :) Would you recommend something better?

src/Components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
+import styled from 'styled-components'
+import { Trash } from '@styled-icons/ionicons-outline'
+import { Edit } from '@styled-icons/boxicons-solid'
+import { useRandomAvatar } from '../../CustomHooks/useRandomAvatar'
+import { PropsType } from './Card.interfaces'
+
+/*---> Components <---*/
```
marcin-piela-sf, line 7: What is the reason for adding such comments to the code? Could you elaborate on what types of comments do you recommend in the code?

src/Components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
...
+  )
+}
+
+/*---> Styles <---*/
+const CardWrapper = styled.div`
+  width: 400px;
+  height: 336px;
+  background: #ffffff;
+  border-radius: 8px;
+  margin-bottom: 64px;
```
marcin-piela-sf, line 39: Looks like this value is repeated in a different places, how would you improve your styles?

src/Components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
...
+  line-height: 26px;
+  margin-top: 28px;
+`
+
+const Description = styled.p`
+  line-height: 20px;
+  margin-top: 8px;
+`
+
+export default Card
```
marcin-piela-sf, line 92: Did you find it useful to use `default exports`? Do you know any disadvantages of it?

src/Components/DeleteModal/DeleteModal.interfaces.ts
```diff
@@ -0,0 +1,10 @@
+import React, { SetStateAction } from 'react'
+import { CardType } from '../../utils/Shared.interfaces'
+
+export interface PropsType {
+  setShowDeleteModal: React.Dispatch<SetStateAction<boolean>>
+  activeCard: CardType | null
+  setActiveCard: React.Dispatch<SetStateAction<CardType | null>>
+  limit: number
+  filterTerm: string
```
marcin-piela-sf, line 9: Why `DeleteModal` requires `filterTerm`, `limit` props? Looks like this component should have `user` or even `userId` as a prop and it should be enough. What do you think?

src/Components/EditModal/EditModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+  align-items: center;
+`
+
+const MapBox = styled.div`
+  width: 256px;
+  height: 87px;
+  background: rgba(255, 255, 255, 0.6);
+  border-radius: 8px;
+  font-size: 18px;
+  line-height: 23px;
```
marcin-piela-sf, line 163: Do you prefer `px` over `rem`?  Could you please explain your decision?

src/Components/ModalOverlay/ModalOverlay.tsx
```diff
@@ -0,0 +1,22 @@
+import styled from 'styled-components'
+import { PropsType } from './ModalOverlay.interfaces'
+
+/*---> Component <---*/
+const ModalOverlay = ({ children }: PropsType) => {
+  return <Wrapper data-cy='modal-overylay'>{children}</Wrapper>
+}
+
+/*---> Styles <---*/
+export const Wrapper = styled.div`
```
marcin-piela-sf, line 10: Did you find it useful to add styles in the same file as component code?

src/Components/Card/Card.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: PropsType) => {
+  return (
+    <CardWrapper data-cy='card-item'>
+      <HeaderWrapper>
+        <DeleteIcon onClick={handleDeleteIconClick} data-cy='delete-icon' />
+        <EditIcon onClick={handleEditIconClick} data-cy='edit-icon' />
+      </HeaderWrapper>
+      <AvatarWrapper>
+        <Avatar
+          src={useRandomAvatar() || '/avatar2.jpg'}
```
marcin-piela-sf, line 22: `useRandomAvatar` is a react hook, shouldn't it be used before `render`?

marcin-piela-sf (commented):
Thanks, @andersonleal1805 for your submission :) Good job 👍

Please don't change your code, just answer the questions.

Cheers :)

AhmedTohamy01:
> Thanks, @andersonleal1805 for your submission :) Good job 👍
> 
> I left some questions. Looking forward to your replies 😃
> 
> Please don't change your code, just answer the questions.
> 
> Cheers :)

Hi @marcin-piela-sf  :)

Thank you so much for taking time in reviewing this PR.

I will be happy to answer all your questions. Very good questions , I like it :)

Actually, i'm still in my new year holiday (Happy New Year) :) . I will get back to Cairo, Egypt after 2-3 days and will answer all your questions. Is it okay?

marcin-piela-sf:
@AhmedTohamy01 yes sure :) Enjoy your time :)

cypress.json
AhmedTohamy01: @marcin-piela-sf 

Well, Frankly, it's not by mistake, if it's by mistake you not find my ignoring them in the `cypress.json` file :)

The benefit of these files is to give us live examples of test scenarios when we need to know how something can be done. I prefer to keep it in the repo so I can find it easily and just ignore it in the `cypress.json` file.

Also, it's a good opportunity to know how to ignore some test files if you need to ignore them. But I'm okay if we don't want to keep it in a production repo :)

src/GraphQLQueries/GraphQLQueries.tsx
AhmedTohamy01: @marcin-piela-sf 

Well, in order to do this you will need to change the "GraphQLSchema" in this file >> `server/GraphQLSchema/GraphQLSchema.js. `

Replace line 39 which is currently like this >> 
`return User.find({ name: { $regex: args.filter, $options: 'i' } }).limit(parseInt(args.limit))
`

With the below code >>
```
return User.find({ 
            $or: [
              { name: { $regex: args.filter, $options: 'i' } }, 
              { description: { $regex: args.filter, $options: 'i' } }, 
              { address: { $regex: args.filter, $options: 'i' } }
            ] 
          }).limit(parseInt(args.limit))
```

This will allow you to search either by name, address, or description :)

src/Pages/MainPage.tsx
AhmedTohamy01: @marcin-piela-sf 

You are right, it's "any".

How we can fix it ?

1) We add a new interface called "DataType" in `src/utils/Shared.interfaces.ts.`

2) The "DataType" interface should be like this:
```
interface DataType {
users: [{ 
	id: string; 
	name: string; 
	address: string; 
	description: string 
    }]
}
```

3) We import "DataType" interface in `MainPage.tsx` and use it in the `useQuery` hook in line 31 like this:

```
const { loading, error, data } = useQuery<DataType | {users: []}>(GET_ALL_USERS, {
    variables: { filter: filterTerm, limit: limit },
  })
```

src/Pages/MainPage.tsx
AhmedTohamy01: @marcin-piela-sf 

Good question:)

This effect is what makes the search looks "smart", this is what makes the app start searching only after the user stops typing in the search box. 

From a UX perspective the user starts searching for a name, lets say "ahmed", he starts typing "a", then "h", then "m" .. what does the user want? does he want the search to start after he types each letter or does he want the search to start after he finishes typing the whole name? 

The user wants the search to start after he finishes typing what he wants to type, he doesn't want the search to happen when he types each and every letter (maybe this even will annoy him). making the search starts if the user stopped typing for 1 sec gives the user a feeling that this search is really smart and knows when he stops typing and makes a better user experience.

src/Pages/MainPage.tsx
AhmedTohamy01: @marcin-piela-sf 

Well, Generally, in terms of UX, it depends on how long the user will wait. What I see suitable is:

1) If the user will wait longer than 2 seconds a looped animation will be okay.
2) if the user will wait longer than 10 seconds we should use something like a percent indicator where the user can expect when this process will finish.

In this particular project, I think the current looped animation is the best option. (I'm open to any other ideas if you have :) )

src/Components/Card/Card.tsx
AhmedTohamy01: @marcin-piela-sf 

Well, when I saw this question I said "Are there is really disadvantages of default exports?" :)

After I searched about it I found people talking about these problems:

- Inconsistency in naming files.
- your teammates import the components with different names.
- you find it tricky in refactoring.

Generally,  it's about the consistency of naming the components, while "named export" enforces your teammates to know the real name of the component and use it.

Why I don't use "named export"?

Because frankly, I didn't have this problem that needs this solution before :)

All my teammates were good guys who respect the component names and usually we cover this point by our Code Style Guide which we are all forced to follow, 

It's something like this, 
- use only arrow functions, not regular functions.
- the file name should match the component name and we should use it as is in importing. 
- ...... 

But I understand that this may be a problem in other teams and maybe it will be better to always stick to "named export', i don't mind using it  :)

src/Components/DeleteModal/DeleteModal.interfaces.ts
AhmedTohamy01: @marcin-piela-sf 

Well, The delete card process itself doesn't need the "filterTerm" and "limit " props, the purpose of "filterTerm" and "limit" props here is to be used in "refetchQueries". This is what makes the list immediately get updated when you delete a card and the deleted card gets removed immediately from the list.

For example, If there are 3 cards in the list "Ahmed", "David", "Mark". If you deleted "Ahmed" card, you will find that "Ahmed" card disappear immediately from the list because we used "refetchQueries" after we deleted "Ahmed" card, if we didn't use "refetchQueries" you will need to refresh the page in order to not see "Ahmed" card in the list.

src/Components/EditModal/EditModal.tsx
AhmedTohamy01: @marcin-piela-sf 

I don't prefer px over rem and i don't prefer rem over px. for me, it's just tools to do the job :)

I pick the easiest tool that is easy to be understood for all other developers who will work on this code later. 

Another point, i don't use another tool unless it really solves a problem that my current tool can NOT solve.

And frankly, i didn't find any problem that requires me to use rem here to solve it :)

Maybe in the future or in other projects i find myself needing to use rem to solve the problem, and i don't mind using it , but i still didn't find the problem that needs this solution, at least in this project :)

src/Components/ModalOverlay/ModalOverlay.tsx
AhmedTohamy01: @marcin-piela-sf 

Good question :) ​

Well, you may find this a little bit weird if you think in it like the normal CSS code, but if you think in it from the Styled-Component perspective you will find that this is the normal way.

It's like what the name says, it's a "Styled Component", you create the component that has its own styles "stick" to it and ONLY to it, and then you use it anywhere. That's why I added the "Styled Component" in the "Component file" because the Component file is the place where you put the components and this is a component, right?

You may see some code patterns where people put the "Styled Components" in a separate file and import the "Ready Styled components" in the component file and use it. Both are valid approaches in styled-components, and I'm open to using this way if this is required in any project. 

Styled components simply eliminate all the normal CSS\SASS\SCSS problems and extend the CSS abilities to have dynamic styles. 

- You don't need to worry about the unique names of class names.
- You don't need to worry about modifying styles in one class and it may affect other components that use this class also and introduce a bug. 
- You don't need to worry about the orphan, old, unused CSS code that stays in your codebase forever and doesn't do anything.
- You don't need to worry about a complicated network of CSS styles that a lot of components use it and it's is all connected together and if you touch it everything gets messy.

src/Components/Card/Card.tsx
AhmedTohamy01: @marcin-piela-sf 

Well, If I understand what you mean here correctly, You mean we should only call hooks at the top-level, right?

You are totally right 👍 

But this is a Custom Hook, It's just a function that calls other Official React Hooks, and these Official React Hooks are called at the top-level inside this Custom Hook function.

That's why you see it works fine here and doesn't throw any errors,  if we are calling it in the wrong place React will throw an error and will prevent us from doing this, make sense? :)

This is from React Official Docs >>
Custom Hooks are more of a convention than a feature. If a function’s name starts with ”use” and it calls other Hooks, we say it is a custom Hook. The useSomething naming convention is how our linter plugin is able to find bugs in the code using Hooks.

AhmedTohamy01:
@marcin-piela-sf 

Hi Marcin, I found a time to answer all your questions, hope you like the answers :)

See you next Monday at the meeting , have a great weekend :)

PR #55, opened 2021-12-28
Juan Jesus Garcia - Frontend

juangl:
Live Version: https://test-web-stack.vercel.app/
README: https://github.com/juangl/test-web-stack

TODO:

- [x] GraphQL Backend
- [x] Apollo configuration 
- [x] Pagination & Filtering
- [x] Fix relay pagination (!)
- [x] Fix not working apollo cache
- [x] Add README with: Deployment Documentation, Architecture Overview, and Documentation on how to create new users
- [ ] E2E and unit tests using Cypress and react-testing-library


juangl:
https://user-images.githubusercontent.com/1887029/147615987-59cca91b-ea6f-4b3f-a552-cd94b8211506.mp4

styles/globals.css
```diff
@@ -0,0 +1,120 @@
...
+  text-decoration: none;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+h1 {
+  font-weight: 300;
+  font-size: 48px;
```
mat-sf, line 28: What unit do you usually use for font-sizes in your projects, and why?

styles/globals.css
```diff
@@ -0,0 +1,120 @@
...
+  width: var(--container-width);
+  margin: 0 auto;
+}
+
+input {
+  font-family: var(--primary-font-family);
+  font-size: 24px;
+  font-weight: 400;
+  padding: 16px;
+  border: 1px solid rgba(0, 0, 0, 0.1);
```
mat-sf, line 57: That color is used few times, how would you improve it?

styles/userCard.module.css
```diff
@@ -0,0 +1,74 @@
+.card {
+  width: 400px;
```
mat-sf, line 2: How would you make that card more responsive on mobile without just setting pixels (assuming it doesn't have to be pixel perfect)?

components/editUserInfoForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <div className="form-group">
+        <label htmlFor="edit-address-address">Address</label>
+        <input
+          required
+          type="text"
+          name="edit-address-address"
+          {...register("address", { required: true })}
+        />
+      </div>
+      <div className="form-group">
```
mat-sf, line 40: Some of the code here is duplicated, how would you improve it?

package.json
```diff
@@ -0,0 +1,40 @@
...
+    "node-fetch": "^3.1.0",
+    "prettier": "^2.5.1",
+    "react": "17.0.2",
+    "react-dom": "17.0.2",
+    "react-hook-form": "^7.22.5",
+    "react-transition-group": "^4.4.2",
+    "unsplash-js": "^7.0.15",
+    "uuid": "^8.3.2"
+  },
+  "devDependencies": {
```
mat-sf, line 34: Do you have any strategy how you separated dev dependencies vs dependencies?

components/editUserInfoModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+export interface EditUserInfoModalProps {
+  data: UserPayload;
+  onRequestClose: () => void;
+}
+
+export default function EditUserModal(props: EditUserInfoModalProps) {
+  const [updateUser] = useMutation(INCREMENT_COUNTER);
+
+  return (
+    <div className={styles.editModalContainer}>
```
mat-sf, line 38: WDYT about creating a separate Modal component to separate the logic (so modal component can be reusable and tested separately)?

styles/editUserInfoModal.module.css
```diff
@@ -0,0 +1,41 @@
+.editModalContainer {
```
mat-sf, line 1: Can you please elaborate more on why you have all the styles for components in one folder? Advantages/disadvantages?

mat-sf (commented):
Hi Juan! Thank you for the submission, we appreciate the time you spend on this.
I added some comments in the PR so we can have a little discussion.
Besides that I have a few general questions:
1. Why did you decide to use next.js in this project?
1. Have you ever worked with storybook? WDYT about this tool?
1. Do you have experience in writing unit/e2e tests? If yes, what libraries do you use and how do you test components?

If you have any questions, feel free to ask here/send us an email 🙂

styles/globals.css
juangl: If you want to allow the user to change the font size of your website through the browser preferences, then `rem` is the way to go. `rem` represents the `html` element font size and is equal to `16px` on most browsers by default. It has a more intuitive behavior than `em`.

styles/globals.css
juangl: ouh, it should probably be a CSS custom property or a SASS variable. The former is better if you want the ability to override the value at runtime.

components/editUserInfoForm.tsx
juangl: humm, there may be an abstract `<FormInput />` component, this component can receive `name`, `RegisterOptions `, etc. as props, and it can use `useFormContext()` from `react-hook-form` to read from the context the register function and do the register process internally.

Depending on the use case this may or may not be a good idea, e.g. if you want to use this component outside of the react-hook-form context, then this component should be written in a more library agnostic way, and then using something like `<Controller />` to use with react-hook-form.

package.json
juangl: I feel like dependencies/devDependencies are more relevant when writing libraries or backend instead of FE because when you compile something with e.g. Webpack, your bundle becomes standalone. For other use-cases, the strategy may be:

dependencies:
- things needed at runtime
- things that run on the CI, since some CI may not install devDependencies

DevDependencies:
- Things that go away at compile time, e.g. babel, Webpack, Sass.
- things that help to verify code quality that won't run on the CI, e.g. prettier

components/editUserInfoModal.tsx
juangl: yeah, that sounds like a good idea and something you normally do for any medium+ size application.

styles/editUserInfoModal.module.css
juangl: Actually, now that you mention it, I much prefer the "by domain" directory structure instead of "by type", that mean, this:

```
├─ components/
│  ├─ componentA/
│  │  ├─ componentA.module.css
│  │  ├─ componentA.tsx
│  ├─ componentB/
│  │  ├─ componentB.module.css
│  │  ├─ componentB.tsx
│  ├─ componentC/
│  │  ├─ componentC.module.css
│  │  ├─ componentC.tsx
```

over:

```
├─ styles/
│  ├─ componentA.module.css
│  ├─ componentB.module.css
│  ├─ componentB.module.css
├─ components/
│  ├─ componentA.tsx
│  ├─ componentB.tsx
│  ├─ componentC.tsx
```

the reason is that when files are grouped by type, those directories become huge. When files are grouped by domain, you end up with more directories but with fewer files in them.

juangl:
hey, @mat-sf! To answer your questions:

> Why did you decide to use next.js in this project?

I decided to use Next.js to play a little bit with it since I haven't done Next.js for a while, but there are a few benefits that I think are worth highlighting:

- Next.js offers a few optimizations out of the box, some of them are:
  - SSR that will improve the time for First Contentful Paint, data fetching for the first render occurs on the server
  - automatic code-splitting for routes
  - the Image component with a few optimizations like lazy loading and serving images with the right dimensions and modern formats
- Next.js allowed me to build the GraphQL API as a serverless function with no extra setup required
- since Vercel uses edge computing, the SSR and the GraphQL API run on a global network of servers similar to CDN.

> Have you ever worked with storybook? WDYT about this tool?

I have used Storybook, the main benefit IMO is when you work with bigger teams, it's a great way to manage the design system and allow the team to communicate what components you can use and how you use them. Also, sometimes it's handy to use it as an environment to work on individual components in isolation.

> Do you have experience in writing unit/e2e tests? If yes, what libraries do you use and how do you test components?

**Unit testing**

For unit testing components, the library that I use the most is `react-testing-library`. it requires running on a JSDom environment, a spec compliment DOM implementation for Node.js. This is the default environment in Jest. This tool actually renders the component as DOM elements and provides helpers to assert that the right thing was rendered. You can mock other components to make sure that you are testing your component in isolation. 

Cypress also provides the ability to test React components individually: https://docs.cypress.io/guides/component-testing/introduction seems more appropriate for integration testing, haven't used this tool myself.

**E2E**

Cypress is the tool I use for end-to-end testing which is a better tool than other popular solutions like selenium in many aspects. The Cypress tests run on the same process as the application does, unlike selenium that runs tests outside the browser and communicates both worlds asynchronously.

PR #57, opened 2022-01-11
Paulo Basto - Frontend

arakno:
Hi, forgot to send the PR


src/app/Panel.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type Props = {
+    title: string
+    url: string
+}
+
+const Panel = ({...props}: Props) => (
+    <div {...props} id="panel" className="users-panel">
+      <ul>
+        <li className="frame">
```
surbina-sf, line 11: How could you avoid repeating all these components?

src/app/App.tsx
```diff
@@ -0,0 +1,36 @@
+import React, {Component} from 'react';
+import Header from './Header'
+import Body from './Body'
+
+type State = {
+  userId: number
+}
+
+class App extends Component<{}, State> {
```
surbina-sf, line 9: What was the reason to use a class component instead of a function component here?

src/app/Button.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+type Props = {
+    text: string,
+    className: string,
+    onClick: () => void
+}
+
+const Button  = ({text, ...props}: Props)=>{
```
surbina-sf, line 9: This component is not adding any special behavior on top of the `button` label. Why did you decide to create it?

src/app/App.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }))
+  }
+
+  render(): JSX.Element {
+    const {userId} = this.state
+
+    return (
+      <div id="body">
+        <Header filterUser={this.filterUser} />
+        <Body value={userId} />
```
surbina-sf, line 30: This instance of `Body` is passing `value` a prop. However, the component props are defined [here](https://github.com/Superformula/test-web-stack/pull/57/files#diff-8f613a8f21ce6a431aaded9694e9f3c53b4f57cd67e178f5fce30fe181a14439R5) and do not include a `value` prop.

How could you avoid these kind of issues?

surbina-sf (commented):
Thanks, @arakno  for your submission :) Good job 👍

Please don't change your code, just answer the questions.

Cheers :)

src/app/Panel.tsx
arakno: passing data as props and iterating each instance with a loop (not forgetting to key index for optimal performance)

src/app/App.tsx
arakno: Thinking ahead, for filtering purposes I wanted to use component state and lifecycle functions, not able to complete on time. 
Besides this component acts as the container which receives data and propagates downwards.

src/app/Button.tsx
arakno: Because thinking ahead, buttons will definitely be reusable in 99% of the applications, only changing slighlty in color variations, size and content.

src/app/App.tsx
arakno: I had some issues there which I couldn't resolve in time, but guess could get away with assigning :any as type

arakno:
Hi Surbina,

Thank you for taking the time to review, I left my comments on the PR.

Have a great day,
Paulo

<webdeveloper>
    Name:  Paulo Basto
    Email:  ***@***.***
    Web:    www.arakno.net
    Twitter: @arakno
    Skype:  arakno.net
    Mobile: [+44] (0)7404597489
</webdeveloper>

PR #63, opened 2022-02-23
Front End Develop Test Solution

TestThrowaway2022SF:
My solution to the front end developer test for Superformula


src/components/atoms/Button/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+import {
+  StyledButton
+} from "./style";
+
+const Button = (props: ButtonProps) => {
+  const {
+    children, onClick,
+    disabled = false,
+    type = ButtonType.Primary,
+  } = props;
```
omer-sf, line 17: how do you feel about destructuring function arguments rather than inside the function body?

src/components/atoms/Button/index.d.ts
```diff
@@ -0,0 +1,12 @@
+
+export enum ButtonType {
+  Primary = 'primary',
+  Secondary = 'secondary'
+}
+
+export interface ButtonProps {
+  children: string;
+  onClick: (any?) => any;
```
omer-sf, line 9: do click handlers usually return values?

src/components/atoms/Button/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const handleClick = async () => {
+    await onClick();
```
omer-sf, line 20: `onClick` is not async, do we need `await` here?

src/components/molecules/AddressMap/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+import { AddressMapProps } from "./index.d";
+
+import {
+  StyledAddressMapPending,
+  StyledAddressMapWrapper
+} from "./style";
+
+
+const MAPBOX_TOKEN = 'pk.eyJ1IjoidGhldG91cmVyIiwiYSI6ImNqaWNxOGdudTAzbHUzd3JvM25xc3R1cWUifQ.9WThxGQjgxI-mdW1XpFydQ';
```
omer-sf, line 12: should we extract these to environment variables?

omer-sf (commented):
Nick, thank you for your time and effort for the submission. Please find my comments in the PR and do not change your code but instead simply reply with your comments when you have the time.

In addition to the comments, I wanted to point out an issue with the edit form below. Where is the issue in the code and how it could be improved please?

https://user-images.githubusercontent.com/81270898/155634840-39df4932-880c-4e3e-b910-45502a7a1d0d.mp4

Again thank you for your time and submission, we appreciate the effort. Looking forward to your replies.

src/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (!rawUsersData) {
+      FauxAPI.setUsers(defaultUsers);
+    }
+
+    setReady(true);
+  }, [setReady]);
+
+  return (
+    <ModalContext.Provider value={modalConfig}>
+      <div>
```
martinrodriguez-sf, line 32: Is this `div` tag serving any purpose?

src/views/UserListView/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        .toLowerCase()
+        .includes(searchTerm.toLowerCase());
+    });
+  }
+
+  return (
+    <div>
+      <BasicLayout
+        header={header}
+      >
```
martinrodriguez-sf, line 84: Typing anything in the search bar re-renders the entire page. How could we prevent this?

src/views/UserListView/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+} from "./style";
+
+const UserListView = () => {
+  const userState = useHookState<UserModel[]>([]);
+  const [searchTerm, setSearchTerm] = useState<string>();
+
+  useEffect(() => {
+    const userData = FauxAPI.getUsers();
+    userState.set(userData);
+  }, []);
```
martinrodriguez-sf, line 33: `userState` is missing in the dependencies array. 
Is this intentional?
Any suggestions on how to prevent this from happening?

src/data/users.ts
```diff
@@ -0,0 +1,42 @@
+import UserModel from "model/UserModel";
+
+const defaultUsers: UserModel[] = [
```
martinrodriguez-sf, line 3: User list is an array and not an object.
Can you list some pros and cons of each alternative? 
If you change your mind and change the type to use an object, which key would you use to store each user and why?

src/components/organisms/UserCard/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return Math.floor(Math.random()*100);
+};
+
+
+const UserCard = (props: UserCardProps) => {
+  const { user, onUpdate } = props;
+
+  const modalContext = useContext(ModalContext);
+
+  console.log(user.createdAt);
```
martinrodriguez-sf, line 33: How can we avoid this`console.log` before committing code?

src/components/organisms/UserCard/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+      />
+    );
+  }
+
+  return (
+    <StyledUserCard allowHover={true}>
+      <StyledUserCardContent>
+        <StyledUserCardAvatar>
+          <img
+            src={`https://source.unsplash.com/random/300x300?sig=${generateRandom()}`}
```
martinrodriguez-sf, line 58: As you mentioned in the FUTURE_PLANS.md file, the image is different every re-render of this component. 
How would you prevent this from happening?

src/components/organisms/UserCard/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import UserModel, { UserModelUpdate } from "model/UserModel";
+
+
+export interface UserCardProps {
+  user: UserModel;
+  onUpdate: (id: string, data: UserModelUpdate) => any;
```
martinrodriguez-sf, line 6: `handleEditUser` function returns void.
Why does `onUpdate` return any?

src/components/organisms/UserCard/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+} from "./style";
+
+
+const generateRandom = () => {
+  return Math.floor(Math.random()*100);
+};
+
+
+const UserCard = (props: UserCardProps) => {
+  const { user, onUpdate } = props;
```
martinrodriguez-sf, line 29: `user` is being repeated 8 times here, how would you destructure it directly from props?

src/components/organisms/EditUser/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+  fireEvent.change(fieldAddress, { target: { value: NEW_ADDR } });
+
+  const fieldDescription = screen.getByTestId('user_field_description');
+  fireEvent.change(fieldDescription, { target: { value: NEW_DESC } });
+
+  const userForm = screen.getByTestId('user_form');
+  fireEvent.submit(userForm);
+
+  await waitFor(() => {
+    expect(handleSubmit).toHaveBeenCalledWith(initialTestUser.id, {
```
martinrodriguez-sf, line 55: Nice usage of `toHaveBeenCalledWith`! Curious about [this](https://github.com/Superformula/test-web-stack/pull/63/files#r814387688) question.

src/components/atoms/TextInput/style.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  &::placeholder {
+    font-family: ${defaultFontFamily};
+    font-style: normal;
+    font-weight: 300;
+    font-size: 24px;
+    line-height: 30px;
+    display: flex;
+    align-items: center;
```
martinrodriguez-sf, line 49: There a few repeated css properties in the placeholder styles. Are those needed?

FUTURE_PLANS.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Improved Accessibility
+
+While parts of the app (form fields and buttons, primarily) have
+support for tab navigation, leveraging `a11y` principles by
+inserting `tabindex` properties along with better CSS focus states for
+generally navigable UI elements would greatly improve the app's
+accessibility. It would also be beneficial to users requiring screen
+readers and other accessibility tools to add `ARIA` values where
+applicable.
```
martinrodriguez-sf, line 33: It's really good that you are thinking about a11y! 

`tabIndex` does not seem to be an issue here, you can tab through cards but you cannot see it because of styling.

I would recommend changing the card list from `div`s to `ul` and `li`s markup, this way screen readers can provide more details about the structure of the app.

There is a mention to "add ARIA values where applicable", could you point the elements that you think need ARIA values?

martinrodriguez-sf (commented):
Thanks for investing time with this project!
As @omer-sf mentioned, we don't expect any extra change from now.
Left some questions to discuss. Nice job!

TestThrowaway2022SF:
Thank you both for the in-depth review feedback, questions and suggestions!

> In addition to the comments, I wanted to point out an issue with the edit form below. Where is the issue in the code and how it could be improved please?

To fix this issue, there are several possible solutions, and here are three of them:
1. **UserEdit form level** - Call the `reset()` function returned from the `useForm` hook (from `react-hook-form`) in the `onCancel` function in `EditUser` to reset the form state back to its default (the unmodified UserModel data) when the user cancels editing. This solution is the optimal one for keeping the form logic self-contained, and useful if the form will be used elsewhere, and needs to be agnostic to whether it will be re-mounted by a parent post-cancel.
2. **Modal hook level** - In the `useModal` hook, add a `useEffect` handler to watch the state of `isOpen` for a change to `false` and clear the `Modal`'s `contentRef` after a short timeout to ensure completion of a smooth closing animation. This solution is the optimal one to ensure that each request to render the content in the `Modal` will start from a fresh mounting—ideally, this would be an optional aspect of the `useModal` hook instance config, as certain uses of the `Modal` may require or benefit from persisting the rendered children regardless of the `isOpen` state.
   - A more advanced and direct version of accomplishing the same goal without relying on a timeout for the animation completion would be to create and return an `onCloseComplete` function from `useModal` to pass to the `onTransitionEnd` handler on `Modal` -> `StyledModalInner`.  This could ensure the content is always cleared when `isOpen === false` and the CSS transform transition completes fully, and also could be useful for custom open/close side effects that rely directly on the `Modal`'s current visibility to the user.
3. **Generic React brute-force re-render** - when rendering the `EditUser` component, doing so with a React `key` based on a value that changes (for example, a concatenation of the user ID and the current date converted to its Unix epoch time representation) will force the component to re-mount from scratch. This is the easiest, but least optimal solution, and definitely a last resort, as it may cause undesired effects depending on parent component re-renders for a reason unrelated to the form. While a possibility, needing to use this method of forcing a fresh re-render is likely indicative of other possible optimizations and/or issues.

In my opinion, the most robust solution would be to do a combination of implementing both 1 and 2(a) to handle all possible scenarios for a `UserEdit` form reset while making the components more self-contained, maintainable and extensible for future use.

src/index.css
```diff
@@ -0,0 +1,8 @@
+body {
+  background-color: #F8F8F8;
+  margin: 0;
+  padding: 0;
+  font-family: Source Sans Pro, "Helvetica Neue", sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
```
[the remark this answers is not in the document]
TestThrowaway2022SF: Because Styled Component's `GlobalStyle` is rendered and injected into the document `<head>` at runtime by React, the styling rules contained therein won't take effect until the app script bundle is loaded and the initial execution takes place. The styles in this traditional CSS (also can be put directly in a `<style>` tag in the document `<head>`) will be loaded and painted before the React app initialization, therefore displaying styling aspects like the background color to the user before the rest of the app loads. This can be useful for React apps that are reliant on DOM outside that which React is responsible for and/or utilizing a transition to initially display the app.

src/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  useEffect(() => {
+    const rawUsersData = window.localStorage.getItem('usersData');
+
+    if (!rawUsersData) {
+      FauxAPI.setUsers(defaultUsers);
+    }
+
+    setReady(true);
+  }, [setReady]);
```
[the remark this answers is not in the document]
TestThrowaway2022SF: Yes it is, as long as the dependency array is still passed to `useEffect` and is empty (signifying the need for only a single initial call the the handler). In this case, I mistakenly accepted the automated suggestion given by my IDE's ESLint suggestions, whereas I should have simply declared an empty dependency array manually when the warning of a possible undeclared dependency was shown.

src/hooks/useModal.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  return {
+    isOpen,
+    content: contentRef.current,
+    show: handleShowModal,
+    close: handleCloseModal
+  };
+};
+
+export default useModal;
```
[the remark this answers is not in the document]
TestThrowaway2022SF: My reasoning for this abstraction was to de-couple the Modal's configuration and state management from the Context provider for added flexibility and extensibility. Keeping the state and its management functions self-contained within this custom hook can allow for simple, scoped addition of further configuration option handling and shared modal behavior beyond what can be done by keeping only the Modal's state in the context, such that these need not be repeated for each component that requires a Modal.

One example of these additions can be found in part of my solution to the `EditUser` form state reset issue, [described here](https://github.com/Superformula/test-web-stack/pull/63#issuecomment-1053901959).

src/App.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <ModalContext.Provider value={modalConfig}>
```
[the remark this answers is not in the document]
TestThrowaway2022SF: To prevent unnecessary re-renders of components that do not use the `ModalContext`'s provided value, it would be better to wrap just the child view rendered by `App` (in this case, `<UserListView />`) in the provider to make sure only the views that will consume the provided context are affected by changes to it.

src/views/UserListView/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  let displayedUsers = users;
+  if (searchTerm) {
+    displayedUsers = users.filter((user) => {
+      return user.name
+        .toLowerCase()
+        .includes(searchTerm.toLowerCase());
+    });
+  }
```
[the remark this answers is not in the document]
TestThrowaway2022SF: In this specific case where the simple existence of a search term is the sole aspect used to decide to filter the users, this suggested ternary is indeed a better and cleaner solution than my current if statement.

If the behavior of the view depended on several other additional conditional aspects of the search term (for instance, preventing search for certain restricted terms, or showing a special UI for certain terms), an if-else if tree (or switch statement) with multiple filtering/rendering paths could be the preferable solution.

src/views/UserListView/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import {
+  StyledUserListContainer,
+  StyledLoadMoreWrapper,
+  StyledUserList,
+  StyledUserListItem
+} from "./style";
+
+const UserListView = () => {
+  const userState = useHookState<UserModel[]>([]);
```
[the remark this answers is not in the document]
TestThrowaway2022SF: React Hookstate as a state management solution is designed to have optimal performance surrounding mutation of nested state (such as my array of UserModel objects), wherein updates can be targeted to specific nested objects / fields without recreating the entire array/state, preventing issues stemming from traditional/native state comparison that may trigger unnecessary re-renders. This would become even more valuable with the implementation of another nesting level (a 2D array representing pages of UserModels for the `Load More` feature).

More information about scoped state in React Hookstate, plus performance measurements, can be found [here](https://hookstate.js.org/docs/scoped-state/).

src/components/organisms/EditUser/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  render(
+    <EditUser
+      user={initialTestUser}
+      onSubmit={handleSubmit}
+      onCancel={handleCancelStub}
+    />
+  );
+
+  const fieldName = screen.getByTestId('user_field_name');
```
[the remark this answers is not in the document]
TestThrowaway2022SF: Yes, certainly. Instead of targeting developer-centric identifiers to test the fields in question, using React Testing Library's queries for label, role, placeholder, text, alt text, etc would be far more user-centric and better than my use of `testid`, which should be used if relying on the others are not possible (which they would have been, in my case). As such, overuse of `getByTestId` may signify user accessibility issues in TDD.

src/views/UserListView/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const handleSearch = (event: SyntheticEvent) => {
+    const newValue = (event.target as HTMLInputElement).value;
+    setSearchTerm(newValue);
+  };
+
+  const userListSearch = (
+    <div>
+      <TextInput
+        onChange={handleSearch}
```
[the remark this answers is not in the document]
TestThrowaway2022SF: There are two possible solutions to improving this.

- The best solution, detailed in my [FUTURE_PLANS.md](https://github.com/TestThrowaway2022SF/test-web-stack/blob/master/FUTURE_PLANS.md#unsplashcom-avatars) file, would have been to set the avatar URL using the Unsplash API at the time when the initial persisted user data is created, thusly tying one avatar permanently to a user.
- The second best solution, assuming no change to the `UserModel` schema or data initialization code, would be to keep a ref or other persistence local to the `UserListView` for all downloaded avatars, each specific to a user ID so the same avatar can be shown for a specific user when the search change re-renders the list.

src/components/atoms/Button/index.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
```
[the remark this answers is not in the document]
TestThrowaway2022SF: It is a trade-off between simpler import paths (`components/atoms/Button` vs `components/atoms/Button/Button`) and component renaming/refactoring vs. better IDE compatibility. I made the choice to use `index` as the root file name for a component because my IDE search is set up to recognize components structured in this way, but in a project with multiple users and a non-standard IDE configuration, named files would likely be a superior solution.

src/components/atoms/Button/index.tsx
TestThrowaway2022SF: It depends on the use case, but I find that destructuring the arguments in the function body and explicitly assigning them to a constant can improve immutability and prevent accidental overwrite of the parameters (props). Attempting to overwrite a const-assigned, body-level destructured parameter will be met with a runtime error, whereas overwriting a function-argument-level parameter would likely only be met with a develop-time warning in some IDEs and dev environments.

src/components/atoms/Button/index.d.ts
TestThrowaway2022SF: This was an oversight on my parent, and the `onClick` handler definition should specify optionally returning a boolean. This allows control (specifically, prevention of default behavior by returning `false`) in cases such as a form submission. This can be used by itself or in conjunction with `event.preventDefault()` from the event passed as a parameter to the `onClick` handler.

src/components/atoms/Button/index.tsx
TestThrowaway2022SF: In this specific case, it does not, and it could be removed. My logic here was to make the `Button` compatible with promises, in the case that a loading animation state may be added to the `Button` to improve user feedback and prevention of extra inputs/clicks.

src/components/molecules/AddressMap/index.tsx
TestThrowaway2022SF: The specific token type used here is a public key token as supplied for front-end use by Mapbox, but could easily be placed in a config file / environment variable, especially in order to differentiate between keys for development, testing and production.

src/App.tsx
TestThrowaway2022SF: No, this `div` can be removed, as it doesn't add anything to the page structure, layout or styling. These aspects are handled within its children (`UserListView`, `Modal`).

src/views/UserListView/index.tsx
TestThrowaway2022SF: The `BasicLayout` wrapper should be moved to a higher component (perhaps `App.tsx`, or its own HOC), while the user list state management and rendering would be a child of it to separate the concerns and state for optimized re-rendering.

src/views/UserListView/index.tsx
TestThrowaway2022SF: Yes, in this case it's an intentional omission, as my logic here was to only call this `useEffect` hook handler once upon the initial view component mounting. My IDE is configured to display a warning when prospective `useEffect` handler dependencies haven't been specified, and in my haste, I actually mistakenly fell victim to this manner of automated ESLint fix suggestion elsewhere in the app, [as explained here](https://github.com/Superformula/test-web-stack/pull/63#discussion_r815592262).

Typically however, these warnings/suggestions can be very useful to aid in composing effective `useEffect` handlers that hinge upon changes to state, refs or other mutable dependencies.

src/data/users.ts
TestThrowaway2022SF: For keeping this user list as an object, I would elect to use the `id` as a key. This is because the user's ID is likely its database identifier—it would be unique, and not mutable by the user, ensuring a valid 1-to-1 relationship between a key and the user it represents in the list.

The pros of an object are direct and instant access to a user by its ID, at the expense of a guaranteed order, and the need to create an array of the object's values (UserModels) to loop through for rendering via a native Object static method (ex: `Object.values(userList)`).

The pros of using an array for the list are a guarantee of the order for rendering and access to native sorting and filtering methods on the Array object, at the expensive of the need to loop through the list to find a specific user (either by ID or another field).

src/components/organisms/UserCard/index.tsx
TestThrowaway2022SF: I should have paid better attention to my environment's ESLint warnings after adding this while troubleshooting an issue, as such a warning against leaving a `console.log` call would have been reported.

src/components/organisms/UserCard/index.tsx
TestThrowaway2022SF: See [this answer](https://github.com/Superformula/test-web-stack/pull/63#discussion_r815610015) for two possible solutions.

src/components/organisms/UserCard/index.d.ts
TestThrowaway2022SF: Similar to the `Button` `onClick` case [explained here](https://github.com/Superformula/test-web-stack/pull/63#discussion_r815617240), in this case, it should return `void` as my usage of it specific to this case has no return. I was considering cases in which a function that has a return maybe used as the callback, however, the better solution in those cases, now that I think about it, would be to wrap functions of that nature so as to prevent a return, before passing the wrapped function as the  `onUpdate`.

src/components/organisms/EditUser/index.test.tsx
TestThrowaway2022SF: Thank you! Question [answered here](https://github.com/Superformula/test-web-stack/pull/63/files#r815607970).

src/components/atoms/TextInput/style.ts
TestThrowaway2022SF: In most browsers, they should not be, but in my past experience I have sometimes seen cases where placeholders use browser-specific styling instead of the custom field styling, and require manual overriding to be styled as desired, so I ensured that styling here accordingly.

src/components/organisms/UserCard/index.tsx
TestThrowaway2022SF: In order to prevent the need to repeat `user` to access its various fields, I could change the line

```
const { user, onUpdate } = props;
```

to this:

```
const {  user, user: { id, name, description, createdAt }, onUpdate } = props;
```

Note that the initial  (non-destructured)  `user` must be specified as well, so as to pass to the `EditUser` component as a consolidated object.

FUTURE_PLANS.md
TestThrowaway2022SF: > I would recommend changing the card list from divs to ul and lis markup, this way screen readers can provide more details about the structure of the app.

The `UserCard` instances rendered by `UserListView` are indeed currently rendered using the `ul` -> `li` list semantics. The `StyledUserList` is a Styled Components-created `ul` tag while the `StyledUserListItem` that directly wraps the `UserCard` is a Styled Components-created `li` tag. I felt these tags were more applicable semantically at the `UserListView` component level rather than the `UserCard` level, as the `UserCard` might end up being used elsewhere in the app—an individual user profile or hover tooltip for a user name, for instance.

> There is a mention to "add ARIA values where applicable", could you point the elements that you think need ARIA values?

Some examples where I would use ARIA values are as follows:
- The `Modal` component's `StyledModalInner` should have the `aria-modal` set to designate it as a modal to the accessibility tree, and `aria-hidden` to denote whether it's currently visible to the user's accessibility device.
- The `FormField` component label should have `aria-label` set to match the form label visibly displayed in the DOM, and the form inputs themselves should have `aria-required` for validation and `aria-disabled` and/or `aria-readonly` for fields that are not currently mutable by the user.
- Likewise, the `FormField` component's `StyledFormFieldError` should have `aria-invalid` and `aria-errormessage` set accordingly to show a user that they have triggered an invalid form state.

TestThrowaway2022SF:
I've done my best to answer all the questions and feedback as needed—please let me know if I missed anything, or if you have any further questions or need elaboration on my answers.

Thank you again for the quality feedback—it's much appreciated for my own learning and research!

PR #64, opened 2022-03-01
Senior Frontend Engineer Test [UI Focused]

sfprojecttestwebstack:
This pull request contains my project.

I've included a deployed version, as well as described a lot inside the README file.

I'm going to send Jordan an email right after I create this with 2 env variables in case you want to run it locally too.

Let me know if you have any questions, but otherwise I am eager to hear your feedback.


omer-sf (commented):
Quinton, thank you for your submission, I know it's a lot to ask and we appreciate the hard work.

I'd like note that I highly enjoyed reviewing your code and my questions are mostly knit picking. Having said that, I do look forward to your reply on the server state management question.

I especially liked where you made sure the fetch would fail at times to make the app more solid.

Wanted to note that there are some mismatches for the spacings from the design and wanted to point that in case you weren't aware. In the edit dialog map vs form mis-alignment being the most obvious.

Please make no changes to the code but reply to the questions when you have the time finalizing the review cycle.

Again, I'd like to thank for your time and look forward to your replies!

PR #65, opened 2022-04-05
React Apollo GraphQL Implementation

romines:
See [README](https://github.com/romines/test-web-stack/blob/master/README.md) for more details!


client/src/components/UserManager/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+  // Handle user search and pagination with query string
+  const [searchParams, setSearchParams] = useSearchParams();
+  const initialQuery = searchParams.get('q') ?? '';
+  const updateUrl = (updatedSearchParams: SearchParams): void => {
+    const current = Object.fromEntries(searchParams.entries()) ?? {};
+    const merged = { ...current, ...updatedSearchParams };
+    setSearchParams(Object.entries(merged));
+  };
+  const page = parseInt(searchParams.get('page') ?? '1');
+  const searchUsersQuery = searchParams.get('q');
```
omer-sf, line 35:
```suggestion
  const searchUsersQuery = searchParams.get('q') ?? {};
```

client/src/components/UserManager/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  // user data graphql query
+  const {
+    data: userData,
+    loading: usersLoading,
+    error: usersError,
+  } = useGetUsersQuery({
+    variables: { page, ...(searchUsersQuery ? { search: searchUsersQuery } : {}) },
```
omer-sf, line 43:
```suggestion
    variables: { page, ...{ search: searchUsersQuery } },
```

client/src/graphql/query.ts
```diff
@@ -0,0 +1,61 @@
+import { gql } from '@apollo/client';
+
+export const GET_USERS = gql`
```
omer-sf, line 3: why are we exporting these? will they ever be utilized? aren't we using a code generator to create types and functions?

client/src/components/UserCard/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (scrollTo && containerRef.current) {
+      containerRef.current.scrollIntoView();
+    }
+  }, [scrollTo]);
+
+  // allow modal launch with enter key
+  const handleEnterKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    if (e.code === 'Enter') {
+      // using DOM APIs feels OK here because it is raw DOM events we care about
+      document.addEventListener('keydown', function handler(e: KeyboardEvent) {
```
omer-sf, line 39: event listener is never removed if we don't press `Escape`. If we close the form and open open the card this event fires twice and will fire more unless we clean up the listener.

client/src/components/UserManager/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+  });
+
+  const users = userData?.users?.users ?? [];
+  const count = userData?.users?.count ?? 0;
```
omer-sf, line 47:
```suggestion
  const count = users.length;
```

client/src/components/UserManager/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  /*
+   *
+   * Local state and user editing
+   *
+   *
+   */
+  const [currentlyEditingUserId, setCurrentlyEditingUserId] = useState<string | null>(null);
+  const [userAddress, setUserAddress] = useState<string | null>(null);
+  const currentlyEditingUser = users.find((u: User) => u.id === currentlyEditingUserId) ?? null;
```
omer-sf, line 74: why the need for `?? null`;

```suggestion
  const currentlyEditingUser = users.find((u: User) => u.id === currentlyEditingUserId);
```

client/src/components/UserManager/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  /*
+   *
+   * Local state and user editing
+   *
+   *
+   */
+  const [currentlyEditingUserId, setCurrentlyEditingUserId] = useState<string | null>(null);
+  const [userAddress, setUserAddress] = useState<string | null>(null);
```
omer-sf, line 73: avoiding `null` and using `undefined` instead would make the rest of the code easier. Also it would help avoid using `as string` in most cases in this component. i.e.:

instead of:
```
const currentlyEditingUser = users.find((u: User) => u.id === currentlyEditingUserId) ?? null
```

simply:
```
const currentlyEditingUser = users.find((u: User) => u.id === currentlyEditingUserId)
```

Instead of: 
```
setUserAddress(currentlyEditingUser?.address as string);
```

simply:
```
setUserAddress(currentlyEditingUser?.address);
```

client/src/App.tsx
```diff
@@ -0,0 +1,21 @@
+import './App.scss';
+
+import { ApolloClient, ApolloProvider, InMemoryCache } from '@apollo/client';
+import { UserManager } from 'components';
+
+const client = new ApolloClient({
+  uri: 'http://localhost:4000/graphql',
```
omer-sf, line 7: we probably should not hard code urls

omer-sf (commented):
Thank you for your submission. It takes a lot of time and effort to complete it and we appreciate all your efforts. 

Overall a very good work, thank you. Please find my comments (mostly nit-picking) here and in the PR. No need to make any changes to the code, but only reply if you have the time please.

* Any reason why partials or modules were not used when styling with Sass? Do you think they can help in larger projects? BEM is usually paired with Sass, have heard about it?

* Do you think usage of `index` files could create difficulty in the IDEs?

* I see type casting used regularly - which I believe should be avoided where possible, defaulting to type safe option. We are essentially telling the compiler we know better of the data than it does - which may not be true at run time and might cause run time errors

client/src/components/UserForm/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  updateUser(u: UserFields): void;
+  setUserAddress: Dispatch<string>;
+}
+
+export default function UserForm({ user, updateUser, setUserAddress, cancel }: IProps) {
+  // grab focus for purposes of keyboard nav
+  const firstInputRef = useRef() as React.MutableRefObject<HTMLInputElement>;
+
+  useEffect(() => {
+    firstInputRef.current && firstInputRef.current.focus();
```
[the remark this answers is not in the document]
romines: Totally agree. Wondering if this is an artifact of some refactor that didn't get cleaned up.

client/src/components/UserForm/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }, []);
+
+  const debouncedAddressChangeHandler = useMemo(() => {
+    const onAddressChangeHandler = (e: React.ChangeEvent<HTMLInputElement>) => {
+      setUserAddress(e.target?.value as string);
```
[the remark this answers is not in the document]
romines: Great suggestion. Nullish coalescing does the best job of enforcing type + communicating intent here.

client/src/components/UserLocation/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  useEffect(() => {
+    if (map) {
+      google.maps.event.clearListeners(map, 'idle');
+
+      if (onIdle) {
+        map.addListener('idle', () => onIdle(map));
+      }
+    }
+  }, [map, onIdle]);
```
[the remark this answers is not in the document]
romines: Because this was taken straight from the official Google Maps example, I didn't scrutinize its use of event listeners closely, but I take that to be the purpose of line 98?

client/src/components/UserLocation/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  if (!deepCompareEqualsForMaps(value, ref.current)) {
+    ref.current = value;
+  }
+
+  return ref.current;
+}
+
+function useDeepCompareEffectForMaps(callback: React.EffectCallback, dependencies: any[]) {
+  useEffect(callback, dependencies.map(useDeepCompareMemoize));
```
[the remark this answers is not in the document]
romines: Likewise, this 'deep comparison' pattern was taken with no modification from the [official Google Maps example](https://github.com/googlemaps/js-samples/tree/main/samples/react-map). I tried to [note](https://github.com/romines/test-web-stack/blob/master/client/src/components/UserLocation/index.tsx#L2) that, but probably should have called more attention to the fact that I deserve no credit (nor blame, hopefully 😎) for most of the contents of this file. The exception is the calls to the Geocoding API on address change.

client/src/components/UserCard/index.tsx
romines: This is a good catch. If the listener was set in a `useEffect` callback, that callback could return a function that removed the listener, which should take care of both the `Escape` key scenario as well as closing the form with the mouse.

client/src/components/UserManager/index.tsx
romines: I can see how this would be misleading to future readers. In this case `count` is the total number of users from the API, whereas `userData.users.users` represents the current 'page' of results in client memory. The ambiguity could probably be resolved with a helpful comment.

client/src/components/UserManager/index.tsx
romines: Good question and I can't think of any 😁

client/src/components/UserManager/index.tsx
romines: 👍

client/src/App.tsx
romines: 👍

client/src/components/UserForm/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface IProps {
+  user: User;
+  cancel(): void;
+  updateUser(u: UserFields): void;
+  setUserAddress: Dispatch<string>;
+}
+
+export default function UserForm({ user, updateUser, setUserAddress, cancel }: IProps) {
+  // grab focus for purposes of keyboard nav
+  const firstInputRef = useRef() as React.MutableRefObject<HTMLInputElement>;
```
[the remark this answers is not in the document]
romines: 👍

romines:
Thanks for substantive feedback! It bodes well for the quality of PR reviews at SF.

w/r/t your top-level comments:

- Any reason why partials or modules were not used when styling with Sass? Do you think they can help in larger projects? BEM is usually paired with Sass, have heard about it?
  - I assumed I would end up using some partials, but in the end had few use cases for shared styles or variables. What little I had ended up in [App.scss](https://github.com/romines/test-web-stack/blob/master/client/src/App.scss)
  - I have used BEM and found it helpful for larger projects. However the syntax is a _little_ ugly/verbose and I thought the scale of this project meant it could be avoided at this stage
- Do you think usage of index files could create difficulty in the IDEs?
  - Difficulty in the sense of ending up with a bunch of tabs that say `index.tsx`? I have a [nice plugin](https://github.com/akinsho/bufferline.nvim) that solves this problem for me, and like being able to do multiple non-default imports on the same line (`import { Modal, UserCard, UserForm } from 'components';`, but if the team had a preference for avoiding index files I would have no problem with that!
- I see type casting used regularly - which I believe should be avoided where possible, defaulting to type safe option. We are essentially telling the compiler we know better of the data than it does - which may not be true at run time and might cause run time errors
  - This is fair, and I think I fell back to this the most around native event handling. I definitely still have some things to learn how React wraps these events and how they are typed!

client/src/components/UserManager/index.tsx
romines: This would work great in conjunction with the suggestion above.

client/src/graphql/query.ts
romines: I had a good reason for doing this at one point, but now I do not remember what that was, and this const is not used anywhere. 😳

PR #66, opened 2022-06-10
Glauber Castro - Frontend code challenge

glauberfc:
Instructions about the project: [README.md](https://github.com/glauberfc/test-web-stack/blob/glauberfc-master/client/README.md)


client/components/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,147 @@
+/** @jsxImportSource @emotion/react */
+import Button from 'components/Base/Button'
+import { Alert, FormGroup } from 'components/Base/Form'
+import { H1 } from 'components/Base/Typography'
+import { Users as User, useUpdateUsersMutation } from 'graphql-files/generated'
+import { MouseEvent } from 'react'
+import { useForm } from 'react-hook-form'
```
nickk-sf, line 7: What is your preferred way of handling forms in React and why? Where do you prefer to keep its state?

client/components/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+    try {
+      await updateUsers({
+        variables: {
+          id: { _eq: user?.id },
+          _set: { ...data },
+        },
+      })
+      onSubmitSuccess()
+    } catch (error) {
+      console.error(error)
```
nickk-sf, line 49: If you were in prod environments, would you handle errors differently? If yes, how?

client/components/EditUserForm/EditUserForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+          <Button type="submit" disabled={isSubmitting}>
+            {isSubmitting ? 'Saving...' : 'Save'}
+          </Button>
+          <Button
+            type="button"
+            variant="secondary"
+            onClick={handleCancelClick}
+            disabled={isSubmitting}
+          >
+            Cancel
```
nickk-sf, line 141: How do you manage and where do you keep static text in prod apps? Any approaches / solutions you tried / can think of>

client/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react'
+import { Container, Overlay } from './Modal.styles'
+
+type ModalProps = {
+  isOpen: boolean
+  children: React.ReactNode
+}
+
+export default function Modal({ isOpen, children }: ModalProps) {
```
nickk-sf, line 9: How would you implement this requirement without using any external solutions? If you can show some code, that would be welcome.

> Appear/Disappear of modal should be animate

client/components/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export default function SearchInput({
+  searchCallback = () => {},
+  clearSearchCallback = () => {},
+}: SearchInputProps) {
+  const [value, setValue] = useState('')
+  const timeoutRef = useRef<NodeJS.Timeout>()
+
+  function handleChange(event: ChangeEvent<HTMLInputElement>) {
```
nickk-sf, line 18: Can you think of any ways to improve performance here? What if the user types too fast meaning that handler will kick in too often? What options do exist to improve perf and user experience?

client/cypress.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'cypress'
```
nickk-sf, line 1: What's your preferred tool to work with e2e tests? Cypress or something different? Why?

client/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "@graphql-codegen/typescript-operations": "^2.4.0",
+    "@graphql-codegen/typescript-react-apollo": "^3.2.14",
+    "deepmerge": "^4.2.2",
+    "dotenv": "^16.0.1",
+    "graphql": "^16.5.0",
+    "lodash": "^4.17.21",
+    "next": "12.1.6",
+    "react": "18.1.0",
+    "react-dom": "18.1.0",
+    "react-hook-form": "^7.31.3",
```
nickk-sf, line 34: What is your experience working with any of these dev libs? 
What is the preferred stack for prod apps you you were asked to start from scratch? Why?

client/pages/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+const Home: NextPage = () => {
+  const router = useRouter()
+  const [modalStatus, setModalStatus] = useState<ModalStatus>({ isOpen: false })
+  const page = Number(router.query?.page)
+  const defaultQueryVariables = {
+    limit: ITEMS_PER_PAGE,
+    offset: (page - 1) * ITEMS_PER_PAGE,
+  }
+
+  const { data, loading, error, fetchMore, refetch, networkStatus } =
```
nickk-sf, line 37: What if the query fails? How would you handle a server error? What is your approach for prod apps you worked on?

nickk-sf (commented):
Hello Glauber! 

Thanks for the time and effort you put into this code challenge. Well done. On top of inline comments I made, I'd also love to learn some more about the following aspects:

1. Do you have experience with a11y? How would yo make this app more accessible?
2. Any prior experience with AWS, and lambdas in particular?

client/components/Base/Avatar.ts
```diff
@@ -0,0 +1,14 @@
+import styled from '@emotion/styled'
```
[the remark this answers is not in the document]
glauberfc: I actually prefer working with Tailwind CSS, but as the team said they "would like to see pure CSS/SASS assets" and I saw Kent C. Dodds using Emotion in React Epic Dev, I decided to give it a try. I liked the flexibility brought by the css prop in the Emotion API

client/components/Base/Form.ts
```diff
@@ -0,0 +1,43 @@
...
+  border: '1px solid rgba(0, 0, 0, 0.1)',
+  background: theme.colors.base,
+})
+
+export const FormGroup = styled.div(({ theme }) => ({
+  display: 'flex',
+  flexDirection: 'column',
+  '> label': {
+    fontWeight: 600,
+    fontSize: '1.125rem',
```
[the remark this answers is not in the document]
glauberfc: I think the main reason is:

> Defining font sizes in px is [not accessible](https://en.wikipedia.org/wiki/Web_accessibility), because the user cannot change the font size in some browsers. For example, users with limited vision may wish to set the font size much larger than the size chosen by a web designer. Avoid using them for font sizes if you wish to create an inclusive design.
[https://developer.mozilla.org/en-US/docs/Web/CSS/font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size)

So since REM is a relative unit, it won't cause this problem.

client/components/EditUserForm/EditUserForm.styles.ts
```diff
@@ -0,0 +1 @@
+export {}
```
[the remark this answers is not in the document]
glauberfc: No, we don't need it. I just tried to show that using this folder structure we can have all files related with an specific component together on the same folder, e.g., `EditUserForm.styles.ts`, `EditUserForm.stories.tsx`, `EditUserForm.test.ts`, etc)

client/components/EditUserForm/EditUserForm.tsx
glauberfc: My preferred way is using `react-hook-form` and when necessary using `yup` for validations. I prefer to keep its state inside the form, handling it by `react-hook-form`.

client/components/SearchInput/SearchInput.tsx
glauberfc: We could turn the input into an uncontrolled component, put it in a form and call the callbacks only when the user hits enter (submit the form) or using an uncontrolled input we could debounce the onChange event, calling the callbacks only after the given debounce time.

client/cypress.config.ts
glauberfc: I prefer using Cypress for e2e testing. I like its simplicity, user experience and documentation. Previously I used Selenium.

client/package.json
glauberfc: In my previous jobs I worked with these libs developing different applications. My preferred stack is Next.js, TypeScript, Tailwind CSS, Apollo client and React Hook Form to handle forms states. 

- **Next.js**: because of features like image optimization, file-system routing, SSG, SSR and SS, TypeScript support, easy integration with Jest, etc;
- **TypeScript**: has become the most popular tool for static type checking in the JavaScript world and has a great community around it;
- **Tailwind CSS**: I really like the dev experience of styling components using it;
- **Apollo client**: it's the GraphQL client I have more experience working with. But I want to learn other tools like React Query/SWR + graphql-request and compare their pros and cons;
- **React Hook Form**: because it's a great tool to handle form state.

client/pages/index.tsx
glauberfc: If the query fails I generally show an alert using a lib like React-Toastify and track the error using a service like Sentry. In some cases I can try to repeat the query automatically or show the user an option to do so.

client/components/Modal/Modal.tsx
glauberfc: The appear state animation is already implemented. To implement the disappearing, I will probably handle the close method asynchronously and add a class to do the closing animation and complete the modal closing when the async method is executed.

glauberfc:
> Hello Glauber!
> 
> Thanks for the time and effort you put into this code challenge. Well done. On top of inline comments I made, I'd also love to learn some more about the following aspects:
> 
> 1. Do you have experience with a11y? How would yo make this app more accessible?
> 2. Any prior experience with AWS, and lambdas in particular?

Thanks for your review!

1. I'm aware of some a11y principles, like "All inputs in a form are associated with a corresponding label element" or "Make sure that all img elements have an alt attribute", but I need to deepen my knowledge about him. One of the things I noticed I did wrong was not using list elements (ol, ul, li) in the user card list. I would follow the [a11y project checklist](https://www.a11yproject.com/checklist) to ensure the accessibility of the app.
2. I've used AppSync and some other AWS services (Cognito, Dynamo DB, etc) for personal projects in the past. I started learning lambdas in the past, but I don't have much experience with it (I'm open to learning though). I have professional experience with Google Cloud Functions which I think is not so different from AWS Lambdas.

client/components/EditUserForm/EditUserForm.tsx
glauberfc: Yes. I would show an alert using a lib like React-Toastify and track the error using a service like Sentry. In some cases I can try to repeat the query automatically or show the user an option to do so.

client/components/EditUserForm/EditUserForm.tsx
glauberfc: I would keep it in a separate file like a `constants.ts`. Or I would keep it in a json file if I'm using a package for internationalization.

